PR #2, opened 2017-06-15
MTA-2699 Changed location of GA override, corrected env variables

richardjamesbaconHMRC:


test/controllers/AddressControllerSpec.scala
```diff
@@ -184,7 +184,7 @@ class AddressControllerSpec extends BaseSpec  {
       val r = c.processTaxCreditsChoice(buildFakeRequestWithAuth("POST").withFormUrlEncodedBody( "value" -> "true" ))
 
       status(r) shouldBe SEE_OTHER
-      redirectLocation(await(r)) shouldBe Some("/tax-credits-service/personal/change-address")
+      redirectLocation(await(r)) shouldBe Some("http://localhost:9242/tax-credits-service/personal/change-address")
```
nicf82: Why is this needed?

test/controllers/PublicControllerSpec.scala
```diff
@@ -78,7 +78,7 @@ class PublicControllerSpec extends BaseSpec  {
 
       val r = c.redirectToExitSurvey(Origin("PERTAX"))(buildFakeRequestWithAuth("GET"))
       status(r) shouldBe SEE_OTHER
-      redirectLocation(r) shouldBe Some("/feedback-survey?origin=PERTAX")
+      redirectLocation(r) shouldBe Some("http://localhost:9514/feedback-survey?origin=PERTAX")
```
nicf82: Why is this needed?

richardjamesbaconHMRC: Prod doesn't have a external-urls section, therefore the exit survey location is blank when it attempts to be read.

On dev, this is needed so it goes to the correct 9514 port. I assume this is the same for the processTaxCreditsChoice logic.

PR #70, opened 2017-10-18
MTA-2730 Use new address when re-selecting an address in the update p…

nicf82:
…ostal address journey

### DIFF
--- a/app/controllers/AddressController.scala
+++ b/app/controllers/AddressController.scala
@@ -86,15 +86,7 @@ class AddressController @Inject() (
   }
 
 
-  //TODO - refactor this logic into addressJourneyEnforcer with boolean enforceNonPostal parameter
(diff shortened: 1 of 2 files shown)


app/controllers/AddressController.scala
```diff
@@ -267,18 +259,23 @@ class AddressController @Inject() (
...
-                    } else {
-                      cacheSelectedAddressRecord(typ, addressRecord) map { _ =>
-                        Redirect(routes.AddressController.showUpdateAddressForm(typ))
-                      }
-                    }
+//                    } else {
+//                      cacheSelectedAddressRecord(typ, addressRecord) map { _ =>
+//                        Redirect(routes.AddressController.showUpdateAddressForm(typ))
+//                      }
+//                    }
```
richardjamesbaconHMRC: Commented lines? Remove?

PR #99, opened 2018-01-12
MTA-2748 changed date formatting for homepage card output for SA tax …

richardjamesbaconHMRC:
…calculations

### DIFF
--- a/app/models/TaxCalculationState.scala
+++ b/app/models/TaxCalculationState.scala
@@ -16,12 +16,13 @@
 
 package models
 
+import java.text.SimpleDateFormat
 import javax.inject.{Inject, Singleton}
 
 import controllers.AddressControllerConfiguration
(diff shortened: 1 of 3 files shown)


app/models/TaxCalculationState.scala
```diff
@@ -85,8 +87,11 @@ class TaxCalculationStateFactory @Inject()(val addressControllerConfiguration: A
     }
   }
 
-  def getSaDeadlineStatus(dueDate: DateTime): Option[SaDeadlineStatus] = {
+  def p302DateToCardDateFormat(p302UnixDate: String): String = {
```
nicf82: Can we put this in DateTimeTools and name it more generacally

richardjamesbaconHMRC: Done

PR #130, opened 2018-02-20
MTA-3142 Added markdown and files for local portal

richardjamesbaconHMRC:


LOCALPORTALREHOMING.mkd
```diff
@@ -0,0 +1,142 @@
+# Portal rehoming environment
```
nicf82: Can we call this file running-local-portal.md

richardjamesbaconHMRC: Done

PR #135, opened 2018-02-23
MTA-3167 Fix bug where SaUtr doesn't show during ambiguous journey

derek-gray:
MTA-3167 Fix bug where SaUtr doesn't show during ambiguous journey


app/controllers/helpers/AmbiguousJourneyHelper.scala
```diff
@@ -0,0 +1,38 @@
...
+import services.SelfAssessmentService
+import uk.gov.hmrc.domain.SaUtr
+
+import scala.concurrent.Future
+
+trait AmbiguousJourneyHelper extends PertaxBaseController with AuthorisedActions {
+
+  def selfAssessmentService: SelfAssessmentService
+
+  def enforceAmbiguousUser(block: SaUtr => Future[Result])(implicit context: PertaxContext): Future[Result] = {
```
nicf82: Any reason this doesnt just go in the Controller?

PR #220, opened 2018-11-21
changed to feature toggle

richardjamesbaconHMRC:


test/controllers/AmbiguousJourneyControllerSpec.scala
```diff
@@ -62,6 +60,7 @@ class AmbiguousJourneyControllerSpec extends BaseSpec {
       def nino: Nino
       def personDetailsResponse: PersonDetailsResponse
       def getSelfAssessmentServiceResponse: SelfAssessmentUserType
+      def saSkipPage : Boolean
```
kurtlogan: `false` is used a lot as a default value, you could add that here and remove it from other tests

app/config/ConfigDecorator.scala
```diff
@@ -167,6 +167,7 @@ class ConfigDecorator @Inject() (configuration: Configuration, langs: Langs) ext
   lazy val allowSaPreview = configuration.getString("feature.allow-sa-preview.enabled").getOrElse("false").toBoolean
   lazy val taxCreditsPaymentLinkEnabled = configuration.getString("feature.tax-credits-payment-link.enabled").getOrElse("true").toBoolean
   lazy val saveNiLetterAsPdfLinkEnabled = configuration.getString("feature.save-ni-letter-as-pdf.enabled").getOrElse("false").toBoolean
+  lazy val saAmbigSkipUTRLetterEnabled = configuration.getString("feature.sa-skip-utr-letter.enabled").getOrElse("true").toBoolean
```
kurtlogan: I would add the property into the `application.conf` file as well just to make it explicit

PR #222, opened 2018-11-22
changed 'register' to 'sign up' in messages files and templates'

richardjamesbaconHMRC:


app/views/ambiguousjourney/usedUtrToEnrolChoice.scala.html
```diff
@@ -19,7 +19,7 @@
 @import uk.gov.hmrc.play.views.html.helpers._
 @import tags._
 
-@main(title = Messages("label.have_you_used_your_utr_to_register"),
+@main(title = Messages("label.have_you_used_your_utr_to_sign_up"),
```
kurtlogan: `enrol` is the official term HMRC uses for this action, perhaps use `enrol` in the keys so if the actual language used to describe this action changes again (which people are currently discussing) we would not need to keep changing the key names with it

richardjamesbaconHMRC: Done

PR #249, opened 2019-01-09
Ddcnos 2849

jpatton638:


app/controllers/PaperlessPreferencesController.scala
```diff
@@ -46,13 +46,19 @@ class PaperlessPreferencesController @Inject() (
...
-  def managePreferences = VerifiedAction(baseBreadcrumb, activeTab = Some(ActiveTabYourAccount)) {
+  def managePreferences: Action[AnyContent] = VerifiedAction(baseBreadcrumb, activeTab = Some(ActiveTabYourAccount)) {
     implicit pertaxContext =>
       showingWarningIfWelsh { implicit pertaxContext =>
         for {
           managePrefsPartial <- preferencesFrontendPartialService.getManagePreferencesPartial(configDecorator.pertaxFrontendHomeUrl, Messages("label.back_to_account_home"))
         } yield {
-          Ok(views.html.preferences.managePrefs(managePrefsPartial.successfulContentOrEmpty))
+          pertaxContext.authProvider match {
+            case Some("IDA") => BadRequest(views.html.error(
```
danewatts: Is it not possible to do this check before calling preferences frontend. This case does not require any data from that service so it makes sense not to bother with the additional call

app/controllers/PaperlessPreferencesController.scala
```diff
@@ -46,13 +46,19 @@ class PaperlessPreferencesController @Inject() (
...
+              s"global.error.BadRequest.title",
+              Some(s"global.error.BadRequest.heading"),
+              Some(s"global.error.BadRequest.message"), showContactHmrc = false))
```
danewatts: s's in these thre messages arent needed

danewatts:
There is no welsh messages added here either

PR #261, opened 2019-02-19
Ddcnls 5285

richardjamesbaconHMRC:
CY-2 taxcalc card


app/config/ConfigDecorator.scala
```diff
@@ -39,6 +39,7 @@ class ConfigDecorator @Inject() (environment: Environment, configuration: Config
 
   val defaultOrigin = Origin("PERTAX")
 
+  //def currentLocalDate: LocalDate = new LocalDate(2019, 1, 1)
```
OktawianC: unused/commented code

conf/application.conf
```diff
@@ -232,7 +232,7 @@ external-url {
     host = "http://localhost:9242"
   }
   taxcalc {
-    host = "http://localhost:9417"
+    host = "http://localhost:9242"
```
OktawianC: I think this should be restored to 9417

richardjamesbaconHMRC: No, 9242 is the pertax-depencencies pport.

OktawianC: OK

app/config/ConfigDecorator.scala
richardjamesbaconHMRC: removed.

PR #263, opened 2019-02-22
SE-121 | Fix bug where country is showing on review changes page

derek-gray:


connor-irvine:
I'll test it on QA where I can test 'postcode look up'

PR #285, opened 2019-06-04
Ddcnls 4128

jpatton638:


test/controllers/SessionManagementControllerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+
+      status(result) shouldBe OK
+    }
+  }
+
+  "SessionManagementController.timeOut" should {
+
+    "return 303" in new LocalSetup {
+
+      val result = controller.timeOut()(FakeRequest("GET", "")).withSession()
```
kurtlogan: `withSession()` not needed

test/controllers/SessionManagementControllerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+
+      status(result) shouldBe SEE_OTHER
+
+    }
+
+    "redirect to the session timeout page" in new LocalSetup {
+
+      val result = controller.timeOut()(FakeRequest("GET", "")).withSession()
```
kurtlogan: `withSession()` not needed

test/controllers/SessionManagementControllerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+
+      redirectLocation(result).getOrElse("Unable to complete") shouldBe routes.PublicController.sessionTimeout().url
+    }
+
+    "clear the session upon redirect" in new LocalSetup {
+
+      val result = controller.timeOut()(FakeRequest("GET", ""))
```
kurtlogan: This test will pass even if `withNewSession()` isn't called in prod code. add `withSession("test" -> "session")`  to the call to fix this. Also don't forget that we should be doing TDD and making sure our tests fail before we write production code

PR #286, opened 2019-06-11
DDCNLS-5842 | Circuit breaker implementation for calling TAI

gorkemunsal:


test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
+
+      val taiService = service(mockSimpleHttp)
+
+      val result1 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result1) shouldBe TaiNotAvailable
+      Thread.sleep(500)
+      val result2 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result2) shouldBe TaiNotAvailable
+
+      Thread.sleep(2000)
```
kurtlogan: 2000ms is way way too long for a unit test. After some tests this will work with 15ms and 30ms. I ran 10000 tests to confirm this and all passed

test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
+
+      val result1 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result1) shouldBe TaiNotAvailable
+
+      Thread.sleep(500)
+
+      val result2 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result2) shouldBe TaiNotAvailable
+
+      Thread.sleep(2000)
```
kurtlogan: Too long see above comment

test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
+
+      Thread.sleep(500)
+      val result2 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result2) shouldBe TaiNotAvailable
+
+      Thread.sleep(500)
+      val result3 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result3) shouldBe TaiNotAvailable
+
+      Thread.sleep(2000)
```
kurtlogan: Too long see above comment

test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
+        .thenReturn(
+          Future.successful(TaxComponentsUnexpectedResponse(httpResponse)),
+          Future.successful(TaxComponentsSuccessResponse(TaxComponents(Seq("EmployerProvidedServices", "PersonalPensionPayments"))))
+        )
+
+      val taiService = service(mockSimpleHttp)
+
+      val result1 = taiService.taxComponents(Fixtures.fakeNino, 2014)
+      await(result1) shouldBe TaiNotAvailable
+      Thread.sleep(500)
```
kurtlogan: Use `.futureValue` instead of `await` and this will mean you do not need to sleep the thread. You will need to mixin `ScalaFutures`

Same for the other sleeps

app/services/TaiService.scala
```diff
@@ -28,53 +29,72 @@ import uk.gov.hmrc.domain.Nino
...
+        ).map { r =>
+          Logger.warn(s"Returned response from TAI $r")
+          r
+        }.collect {
+          case response@(TaxComponentsSuccessResponse(_) | TaxComponentsUnavailableResponse) => response
         }
-      )
+      }
     }
+    circuitBreaker.withCircuitBreaker(call).fallbackTo(Future.successful(TaiNotAvailable))
```
kurtlogan: instead of implementing the circuit breaker direct on the service it may be better to use named dependencies and implement it as it's own implementation of the `TaiService` component. This will help break up the concerns and also make switching between the cb pattern and no cb pattern simple

As an example
```
trait TaiService {
  def taxComponents(nino: Nino, year: Int)(implicit hc: HeaderCarrier): Future[TaxComponentsResponse]
}

@Singleton
class CircuitBreakerTaiService @Inject()(@Named("default") underlying: TaiService, circuitBreaker) extends TaiService {
  def taxComponents(nino: Nino, year: Int)(implicit hc: HeaderCarrier): Future[TaxComponentsResponse] = {
    val call = userlying.taxComponents(nino, year).collect {
       case response@(TaxComponentsSuccessResponse(_) | TaxComponentsUnavailableResponse) => response
    }

    circuitBreaker.withCircuitBreaker(call).fallbackTo(Future.successful(TaiNotAvailable))
  }
}

class DefaultTaiService { ... remains the same... }
```
and also add to `LocalGuiceModule`
```
bind[TaiService].qualifiedWith("default").to[DefaultTaiService]
```

test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
-      val metricId = "get-tax-components"
+      val metricId: Metric = "get-tax-components"
+    }
+
+    "circuitBreaker must call tai" in new LocalSetup {
+      override lazy val simulateTaiServiceIsDown = false
+      val mockSimpleHttp = MockitoSugar.mock[SimpleHttp]
+      override lazy val httpResponse = HttpResponse(SEE_OTHER)
+
+      when(mockSimpleHttp.get[TaxComponentsResponse](Matchers.any())(Matchers.any(), Matchers.any())(Matchers.any()))
```
kurtlogan: You have imported `Matchers` so you can just use `any()`

test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
+      await(result3) shouldBe TaxComponentsSuccessResponse(TaxComponents(Seq("EmployerProvidedServices", "PersonalPensionPayments")))
+
+    }
+
+    "circuitBreaker must call tai when call errors" in new LocalSetup {
+      override lazy val simulateTaiServiceIsDown = false
+      val mockSimpleHttp = MockitoSugar.mock[SimpleHttp]
+      override lazy val httpResponse = HttpResponse(SEE_OTHER)
+
+      when(mockSimpleHttp.get[TaxComponentsResponse](Matchers.any())(Matchers.any(), Matchers.any())(Matchers.any()))
```
kurtlogan: Use `any()`

test/services/TaiServiceSpec.scala
```diff
@@ -16,78 +16,193 @@
...
+      await(result3) shouldBe TaxComponentsSuccessResponse(TaxComponents(Seq("EmployerProvidedServices", "PersonalPensionPayments")))
+    }
+
+    "circuitBreaker must call tai when call sucess, fail then success" in new LocalSetup {
+
+      override lazy val simulateTaiServiceIsDown = false
+      val mockSimpleHttp = MockitoSugar.mock[SimpleHttp]
+      override lazy val httpResponse = HttpResponse(SEE_OTHER)
+
+      when(mockSimpleHttp.get[TaxComponentsResponse](Matchers.any())(Matchers.any(), Matchers.any())(Matchers.any()))
```
kurtlogan: use `any()`

conf/application.conf
```diff
@@ -170,6 +170,20 @@ microservice {
       refreshAfter = 10
       expireAfter = 60
     }
+
+    tai {
```
kurtlogan: There is already a tai section in the config and the port numbers are different. I also can't see where bearer-token in used. Might be better to just include the circuit breaker specific config to the current tai config and  remove the the rest from here

app/config/TaiCircuitBreakerProvider.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  private val maxFailures = config.getInt("microservice.services.tai.circuit-breaker.max-failures")
+    .getOrElse(throw new IllegalStateException("tai.circuit-breaker.max-failures config value not set"))
+  private val callTimeout = config.getMilliseconds("microservice.services.tai.circuit-breaker.call-timeout")
+    .getOrElse(throw new IllegalStateException("tai.circuit-breaker.call-timeout config value not set")).milliseconds
+  private val resetTimeout = config.getMilliseconds("microservice.services.tai.circuit-breaker.reset-timeout")
+    .getOrElse(throw new IllegalStateException("tai.circuit-breaker.reset-timeout config value not set")).milliseconds
+
+  override def get(): CircuitBreaker =
+    new CircuitBreaker(
```
kurtlogan: Was there not some logging to go here?

PR #302, opened 2019-07-24
Ddcnos 5463

andydow311:


danewatts:
Please resolve conflicts before review, @andydow311

danewatts:
Need PR for config to ensure absolute url is not enabled outside of localhost.

PR #303, opened 2019-08-01
DDCNLS-6492 Scotland <-> England address change

oc-ste:


app/views/personaldetails/updateAddressConfirmation.scala.html
```diff
@@ -14,57 +14,64 @@
...
+                        <h1 class="heading-xlarge">@Messages("label.your_postal_address_has_been_changed_")</h1>
+                    </div>
 
-                <p>@Html(addr.get.mkString(", <br>"))</p>
+                    <p>@Messages("label.all_your_letters_will_be_sent_to_this_addr")</p>
 
-                <p><a href="@controllers.routes.HomeController.index" class="button" role="button">@Messages("label.return_to_your_account_home")</a></p>
-            </div>
-        </div>
+                    <p>@Html(addr.get.mkString(", <br>"))</p>
```
kurtlogan: `.map` should actually result in the same result for a `Some` without the issue of throwing an exception
```suggestion
                    <p>@Html(addr.map(_.mkString(", <br>")))</p>
```

app/controllers/AddressController.scala
```diff
@@ -793,48 +796,66 @@ class AddressController @Inject()(
...
+    addressType: String,
+    payeAccount: PayeAccount,
+    personDetails: PersonDetails,
+    address: Address,
+    successResponseBlock: () => Result)(implicit pertaxContext: PertaxContext): Future[Result] =
+    citizenDetailsService.updateAddress(payeAccount.nino, personDetails.etag, address) map {
+
+      case UpdateAddressBadRequestResponse =>
+        BadRequest(
+          views.html.error(
```
kurtlogan: These views and messages seem very generic, is there a better place for them/

oc-ste: Moved these and the UpdateAddressResponse behaviour, along with the tests

app/models/addresslookup/Address.scala
```diff
@@ -60,7 +60,8 @@ case class Address(
   town: Option[String],
   otherLine: Option[String],
   postcode: String,
-  country: Country) {
+  country: Country,
+  subdivision: String) {
```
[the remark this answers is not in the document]
oc-ste: This Address file looks to be copied and pasted from address-reputation-store. The fix should be to just add a dependency on this service but I'm not sure how big of a change it will be. It'll fix a few things in PTA as there is another Country model that looks eerily similar to the one contained in this file.

app/services/AddressMovedService.scala
```diff
@@ -0,0 +1,75 @@
...
+    for {
+      fromResponse <- addressLookupService.lookup(fromAddressId)
+      toResponse   <- addressLookupService.lookup(toAddressId)
+    } yield {
+      (fromResponse, toResponse) match {
+        case (AddressLookupSuccessResponse(fromRecordSet), AddressLookupSuccessResponse(toRecordSet)) =>
+          val fromSub = fromRecordSet.addresses.headOption.map(_.address.subdivision)
+          val toSub = toRecordSet.addresses.headOption.map(_.address.subdivision)
+
+          val addressChanged = fromSub.flatMap(
```
kurtlogan: This assigment no longer needs map and flatMap. You can shorten the block to be
```
case (AddressLookupSuccessResponse(fromRecordSet), AddressLookupSuccessResponse(toRecordSet)) =>
  val fromSub = fromRecordSet.addresses.headOption.flatMap(_.address.subdivision)
  val toSub = toRecordSet.addresses.headOption.flatMap(_.address.subdivision)
          
  if (hasMovedFromScotland(fromSub, toSub))
    MovedFromScotland
  else if (hasMovedToScotland(fromSub, toSub))
    MovedToScotland
  else
    AnyOtherMove
```

PR #305, opened 2019-08-06
Ddcnls 6143

jpatton638:


app/config/ConfigDecorator.scala
```diff
@@ -233,4 +236,30 @@ class ConfigDecorator @Inject()(environment: Environment, configuration: Configu
...
+
+trait TaxcalcUrls {
+  self: ConfigDecorator =>
+
+  def reconciliationsUrl(nino: String, startYear: Int, endYear: Int) =
+    s"${self.taxCalcHost}/$nino/$startYear/$endYear/reconciliations"
+
+  def underpaidUrlReasons(taxYear: Int) =
+    s"${self.taxCalcFrontendHost}/tax-you-paid/$taxYear-${taxYear + 1}/paid-too-little/reasons"
+  def overpaidUrlReasons(taxYear: Int) =
```
oc-ste: Could dedupe $taxYear-${taxYear + 1} with a function

app/controllers/HomeController.scala
```diff
@@ -57,15 +54,11 @@ class HomeController @Inject()(
...
       if (configDecorator.taxcalcEnabled) {
-        taxCalculationService.getTaxCalculation(nino, year) map {
-          case TaxCalculationSuccessResponse(taxCalc) =>
-            Some(taxCalculationStateFactory.buildFromTaxCalculation(Some(taxCalc), includeOverPaidPayments))
-          case _ => None
-        }
+        taxCalculationService.getTaxYearReconciliations(nino, year, year) map (_.headOption)
       } else {
-        Future.successful(Some(TaxCalculationDisabledState(year - 1, year)))
+        Future.successful(None)
```
oc-ste: Missing tests

app/controllers/HomeController.scala
```diff
@@ -76,15 +69,15 @@ class HomeController @Inject()(
...
+      userAndNino.fold[Future[(TaxComponentsState, Option[TaxYearReconciliation], Option[TaxYearReconciliation])]](
         Future.successful((TaxComponentsDisabledState, None, None))) { userAndNino =>
         val (user, nino) = userAndNino
 
         val taxCalculationStateCyMinusOne = getTaxCalculationState(nino, year - 1, includeOverPaidPayments = true)
         val taxCalculationStateCyMinusTwo = if (configDecorator.taxCalcShowCyMinusTwo) {
           getTaxCalculationState(nino, year - 2, includeOverPaidPayments = true)
         } else
-          Future.successful(Some(TaxCalculationUnkownState))
+          Future.successful(None)
```
oc-ste: Think this fold block have missing tests

test/controllers/helpers/HomeCardGeneratorSpec.scala
```diff
@@ -116,144 +119,136 @@ class HomeCardGeneratorSpec extends BaseSpec {
...
     }
 
-    "return nothing when called with TaxCalculationUnkownState" in new LocalSetup {
-      val taxCalcState = TaxCalculationUnkownState
+    "return correct markup when called with Balanced status" in new LocalSetup {
+      val taxYearRec = TaxYearReconciliation(2015, Balanced)
 
-      cardBody shouldBe None
+      cardBody shouldBe taxCalculationViewModel.map(taxCalculation(_))
     }
```
oc-ste: It seems like there are no tests on TaxCalculationViewModel and they are side tested here, would it make sense to push the view model logic into the HomeGenerator (or something else)

app/services/TaxCalculationService.scala
```diff
@@ -72,13 +74,24 @@ class TaxCalculationService @Inject()(
...
+          Logger.debug(e.toString)
+          t.completeTimerAndIncrementFailedCounter()
+          Logger.warn("Error getting tax calculation from tax-calculation-service", e)
+          TaxCalculationErrorResponse(e)
         }
       )
     }
+
+  def getTaxYearReconciliations(nino: Nino, startYear: Int, endYear: Int)(
+    implicit headerCarrier: HeaderCarrier): Future[List[TaxYearReconciliation]] =
```
oc-ste: Missing tests, do we want too track metrics on this similar to getTaxCalculation?

app/viewmodels/TaxCalculationViewModel.scala
```diff
@@ -0,0 +1,259 @@
...
+    case (Overpaid(Some(amount), PaymentPaid), taxYears) =>
+      overpaid("label.hmrc_has_paid_your_refund", amount, taxYears)
+
+    case (Overpaid(Some(amount), ChequeSent), taxYears) =>
+      overpaid("label.hmrc_sent_you_a_cheque_for", amount, taxYears)
+
+  }
+
+  private def overpaid(contentKey: String, amount: Double, taxYears: TaxYears, links: Link*) =
+    TaxCalculationViewModel(
```
oc-ste: Missing return type

app/viewmodels/package.scala
```diff
@@ -0,0 +1,54 @@
+/*
```
oc-ste: Confusing filename, maybe move Heading, Link and TaxYears to TaxCalculationViewModel as it's only used there and the rest to ViewUtils

it/services/TaxCalculationServiceSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+  def hcWithBearer(nino: String): HeaderCarrier = hc.withExtraHeaders(("AUTHORIZATION", getBearerToken(nino)))
+
+  "getTaxYearReconciliations" should {
+
+    "return a successful response" when {
+
+      "data is successfully retrieved for Balanced status" in {
+
+        val result = serviceUnderTest.getTaxYearReconciliations(Nino("JE989013A"), 2017, 2019)(hcWithBearer("JE989013A"))
+
```
oc-ste: Deduple the ninos here being passed into getTaxYearReconciliations and hcWithBearer

test/util/Fixtures.scala
```diff
@@ -102,6 +102,8 @@ trait TaiFixtures {
 
 trait TaxCalculationFixtures {
   def buildTaxCalculation = TaxCalculation("Overpaid", BigDecimal(84.23), 2015, Some("REFUND"), None, None, None)
+  def buildTaxYearReconciliations: List[TaxYearReconciliation] =
+    List(TaxYearReconciliation(2015, Balanced), TaxYearReconciliation(2016, Balanced))
 }
```
oc-ste: This is only used in one place, does it need to be in fixtures?

test/views/html/TaxCalculationViewSpec.scala
```diff
@@ -0,0 +1,340 @@
...
+  import BetterOptionValues._
+
+  def view(reconciliation: Reconciliation): Option[Document] =
+    TaxCalculationViewModel.fromTaxYearReconciliation(TaxYearReconciliation(2017, reconciliation))(config).map {
+      taxRec =>
+        asDocument(taxCalculation(taxRec)(messages, pertaxContext, config).toString)
+    }
+
+  def formatDate(date: LocalDate) = LanguageHelper.langUtils.Dates.formatDate(Some(date), "dd MMMM yyyy")(messages)
+
```
oc-ste: I think a lot of the logic being tested here overlaps with the tests in HomeCardGeneratorSpec, could simplify this spec after the extraction of view logic into one place.

test/views/html/ViewSpec.scala
```diff
@@ -16,14 +16,30 @@
...
 import org.jsoup.Jsoup
 import org.jsoup.nodes.Document
+import org.scalatest.Assertion
+import play.api.i18n.{Lang, Messages}
+import play.api.i18n.MessagesApi
+import play.api.test.FakeRequest
+import util.BaseSpec
 
-trait ViewSpec {
+trait ViewSpec extends BaseSpec {
```
oc-ste: If a ViewSpec is really a BaseSpec then there are a few other places to tidy up where Specs are still extending both ViewSpec with BaseSpec

test/util/Fixtures.scala
jpatton638: This is now used in multiple places

app/services/TaxCalculationService.scala
jpatton638: Discussed this with you

test/controllers/helpers/HomeCardGeneratorSpec.scala
jpatton638: I have looked into this suggestion and believe that it may not be a good idea to move the logic from the View model to the HomeCardGenerator as the generator only serves to return HTML under certain circumstances for each type of tile in Pertax. 

Putting that logic there may make the Generator more convoluted, therefore the decision was made to strip the logic out of the view itself and create a model that create objects with the most appropriate structure, which could then be passed to the view to render in a clear and simple way.

test/views/html/TaxCalculationViewSpec.scala
jpatton638: I would argue that these tests are checking that the correct links and actual content are being rendered as expected whereas the HomeCardGeneratorSpec checks that the function which calls the view behaves as expected by returning HTML given certain pre-conditions and circumstances.

test/controllers/helpers/HomeCardGeneratorSpec.scala
oc-ste: Sorry, I've probably not explained it properly in my first go. 

The tests on TaxCalculationViewModel appear to live in the TaxCalculationViewSpec, I think it would better to place them in a TaxCalculationViewModelSpec and have the ViewSpec just testing that it can print out any link/url. At the moment the ViewModel tests are tied to the ViewSpec tests so the tests are both testing the correct url/labels are being rendered but what you really care about is that the ViewModel creates the right urls/labels and the ViewSpec is capable of rendering any url/label. ViewSpecs should be really light weight as when you start testing logic in a ViewSpec they can become large and unwieldy due to testing logic + rendering simultaneously. We should always be trying to split UI and logic out into separate independent modules where they share no concerns.

test/views/html/TaxCalculationViewSpec.scala
```diff
@@ -0,0 +1,341 @@
...
+  val underpaidUrlReasons: String =
```
kurtlogan: No need to construct these urls from scratch as they already exist in the config decorator. It will also make the system slightly more resilient to change as if a url changes it only needs to change in 1 place

PR #308, opened 2019-08-13
Ddcnls 5809

oc-ste:


app/connectors/PayApiConnector.scala
```diff
@@ -0,0 +1,56 @@
...
+      response.status match {
+        case CREATED =>
+          val createPayment = responseTo[CreatePayment](response)
+          Future.successful(new CreatePaymentSuccess(createPayment))
+        case _ => Future.successful(CreatePaymentFailed)
+      }
+    }
+  }
+
+  private def responseTo[T](response: HttpResponse)(implicit rds: Reads[T]) = response.json.as[T]
```
kurtlogan: `as` will throw an exception if it fails to parse the json

app/connectors/PayApiConnector.scala
```diff
@@ -0,0 +1,56 @@
...
+  def createPayment(
+    request: PaymentRequest)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[CreatePaymentResponse] = {
+    val postUrl = configDecorator.makeAPaymentUrl
+
+    http.POST[PaymentRequest, HttpResponse](postUrl, request) flatMap { response =>
+      response.status match {
+        case CREATED =>
+          val createPayment = responseTo[CreatePayment](response)
+          Future.successful(new CreatePaymentSuccess(createPayment))
+        case _ => Future.successful(CreatePaymentFailed)
```
kurtlogan: Why have you have you chosen to create your own success/failure datatype?
You have a few options available that will give greater usability.

1. Use the future failure channel with a bad request exception
This will allow consumers to not have to consider the failure case of this method as it will automatically short circuit any computations after a failure. The downside is you have to use a `BadRequestException` to have the global error handler interpret it correctly which means this method would need knowledge of how the application is structured as a whole.

2. Use `Either[Error, CreatePayment]` datatype
You get all the good stuff that comes with either like map, fold, flatMap, filter etc which will make the api of the class easier to work with. It also works well when there are multiple failure cases

3. Use `Option[CreatePayment]` datatype
Probably the one I would go with. This can represent the exact same use case that you have implemented here with the added benefit that you get all the good methods that come with option and everyone should know how to deal with them. Of course of there are multiple ways that a method can fail option cannot report on that

test/connectors/PayApiConnectorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  "createPayment" should {
+    "parse the json load for a successful CREATED response" in {
+      val json = Json.obj(
+        "journeyId" -> "exampleJourneyId",
+        "nextUrl"   -> "testNextUrl"
+      )
+
+      when(
+        http.POST[PaymentRequest, HttpResponse](Matchers.eq(postUrl), Matchers.eq(paymentRequest), any())(
```
kurtlogan: Can you import eq? You will need to qualify the name as it will clash, I usually use eqTo
```
import org.mockito.Matchers.{eq=>eqTo}
```

test/connectors/PayApiConnectorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+          any(),
+          any(),
+          any(),
+          any()))
+        .thenReturn(Future.successful(HttpResponse(CREATED, Some(json))))
+
+      val result = Await.result(connector.createPayment(paymentRequest), 5.seconds)
```
kurtlogan: Extend `ScalaFutures` instead and you can use either `.futureValue` or `whenReady(futureHere) { ??? }`
I think it is a cleaner api

test/connectors/PayApiConnectorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+import scala.concurrent.{Await, Future}
+
+class PayApiConnectorSpec extends BaseSpec with MockitoSugar {
+
+  val http = mock[WsAllMethods]
+  val connector = new PayApiConnector(http, config)
+  val paymentRequest = PaymentRequest(config, "some utr")
+  val postUrl = config.makeAPaymentUrl
+
+  "createPayment" should {
```
kurtlogan: You are missing a test here for when invalid json is returned from the api

app/controllers/ApplicationController.scala
```diff
@@ -142,6 +145,20 @@ class ApplicationController @Inject()(
     }
   }
 
+  def makePayment: Action[AnyContent] = VerifiedAction(baseBreadcrumb) { implicit pertaxContext =>
```
kurtlogan: Is this the best place for this method. I'm just think that `ApplicationController` is a very generic name and maybe this would be better placed in a `PaymentsController` or similar

app/connectors/PayApiConnector.scala
```diff
@@ -0,0 +1,56 @@
...
+
+object CreatePayment {
+  implicit val format = Json.format[CreatePayment]
+}
+
+sealed trait CreatePaymentResponse
+final case class CreatePaymentSuccess(createPaymentResponse: CreatePayment) extends CreatePaymentResponse
+object CreatePaymentFailed extends CreatePaymentResponse
+
+class PayApiConnector @Inject()(http: WsAllMethods, configDecorator: ConfigDecorator) {
```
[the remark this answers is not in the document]
oc-ste: Tried this initially, would need to create a concrete HttpPost for this to work, not sure what more this would gain apart from a lightweight HttpPost. Think most of this goes away with the bootstrap upgrade anyway, as they give us a default http client to work with.

app/connectors/PayApiConnector.scala
oc-ste: Tried asOpt but the logging was basically the same as the exception, gone down the route of propagating the exception up. 

Think it was actually ok in a function as it was private but inlined it anyway as it's a one liner.

app/connectors/PayApiConnector.scala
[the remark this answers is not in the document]
kurtlogan: :+1:

PR #309, opened 2019-08-20
DDCNOS-5816 | changing regex same as citizen-details for Address

shikkhasingh:


app/util/PertaxValidators.scala
```diff
@@ -63,7 +63,7 @@ object PertaxValidators {
         value.fold[Map[String, String]](Map.empty)(v => Map(key -> v))
     })
 
-  private val AddressLineRegex = """^[ A-Za-z0-9&’'(),-./sôéàëŵŷáîïâêûü]+""".r
+  private val AddressLineRegex = """^[A-Za-z0-9 \-,.&'\/]+""".r
```
ManoManoharan: Can we replace regex to be  **^[A-Za-z0-9 \\-,.&'\\/]{1,35}$**?

shikkhasingh: 35 character length validation is already there in AddressDto.scala. The reason behind having separate validations for allowed character set and 35 character length is that we need to show two separate error message to the user for respective scenarios.

PR #319, opened 2019-09-05
fixed spacing and two thirds layout

etheya:
Originally doing on only the one page but the National Insurance page also used the same file so changed that out too. Tests all passed.


etheya:
Ive looked at this but can someone just check there is nothing broken - i used the master version and removed the few bits i wanted to remove. But please double check. Havent ran the tests etc.

PR #357, opened 2019-11-12
Revert 356 revert 355 ddcnls 6858

jpatton638:


app/controllers/AddressController.scala
```diff
@@ -331,80 +332,111 @@ class AddressController @Inject()(
...
                        }
             } yield result
           }
         )
       }
     }
 
+  def showAddressSelectorForm(typ: AddrType) =
+    authenticate.async { implicit request =>
+      val postcode =
```
kurtlogan: These two lines are duplicated a few times, consider creating a method like so
```
def postcode(implicit request: Request[_]) = request.body.asFormUrlEncoded.flatMap(_.get(POSTCODE).flatMap(_.headOption)).getOrElse("")
```

app/controllers/AddressController.scala
```diff
@@ -331,80 +332,111 @@ class AddressController @Inject()(
...
   def processAddressSelectorForm(typ: AddrType): Action[AnyContent] =
     authenticate.async { implicit request =>
       val postcode =
-        request.body.asFormUrlEncoded.flatMap(_.get("postcode").flatMap(_.headOption)).getOrElse("")
+        request.body.asFormUrlEncoded.flatMap(_.get(POSTCODE).flatMap(_.headOption)).getOrElse("")
       val filter =
-        request.body.asFormUrlEncoded.flatMap(_.get("filter").flatMap(_.headOption))
+        request.body.asFormUrlEncoded.flatMap(_.get(FILTER).flatMap(_.headOption))
+
+      val errorPage = Future.successful(
```
kurtlogan: This same code is used in  a few places in this file, pull it out into it's own method

app/controllers/AddressController.scala
```diff
@@ -331,80 +332,111 @@ class AddressController @Inject()(
...
-                                case (_, true) => Redirect(routes.AddressController.enterStartDate(typ))
-                                case (_, false) =>
-                                  cacheSubmittedStartDate(typ, DateDto(LocalDate.now()))
-                                  Redirect(routes.AddressController.reviewChanges(typ))
-                              }
+                          val postCodeHasChanged = !postcode
+                            .replace(" ", "")
+                            .equalsIgnoreCase(personDetails.address.flatMap(_.postcode).getOrElse("").replace(" ", ""))
+                          (typ, postCodeHasChanged) match {
+                            case (PostalAddrType, true) => Redirect(routes.AddressController.enterStartDate(typ))
```
kurtlogan: This line can be removed as it is covered in 430

app/controllers/AddressController.scala
```diff
@@ -331,80 +332,111 @@ class AddressController @Inject()(
...
+                            case (PostalAddrType, false) =>
+                              Redirect(routes.AddressController.showUpdateAddressForm(typ))
+                            case (_, true) => Redirect(routes.AddressController.enterStartDate(typ))
+                            case (_, false) =>
+                              cacheSubmittedStartDate(typ, DateDto(LocalDate.now()))
+                              Redirect(routes.AddressController.reviewChanges(typ))
                           }
```
kurtlogan: Replace with a for comprehension as it is cleaner and has less indentation, like so|
```
for {
  _ <- cacheSelectedAddressRecord(typ, addressRecord)
  _ <- cacheSubmittedAddressDto(typ, addressDto)
} yield {
  val postCodeHasChanged = !postcode
    .replace(" ", "")
    .equalsIgnoreCase(personDetails.address.flatMap(_.postcode).getOrElse("").replace(" ", ""))
  (typ, postCodeHasChanged) match {
    case (PostalAddrType, true) => Redirect(routes.AddressController.enterStartDate(typ))
    case (PostalAddrType, false) =>
      Redirect(routes.AddressController.showUpdateAddressForm(typ))
    case (_, true) => Redirect(routes.AddressController.enterStartDate(typ))
    case (_, false) =>
      cacheSubmittedStartDate(typ, DateDto(LocalDate.now()))
      Redirect(routes.AddressController.reviewChanges(typ))
    }
}
```

app/controllers/helpers/AddressJourneyCachingHelper.scala
```diff
@@ -37,6 +37,9 @@ trait AddressJourneyCachingHelper { this: AddressController =>
     implicit hc: HeaderCarrier): Future[CacheMap] =
     sessionCache.cache(s"${typ}SelectedAddressRecord", addressRecord)
 
+  def cacheSelectedRecordSet(typ: AddrType, recordSet: RecordSet)(implicit hc: HeaderCarrier): Future[CacheMap] =
```
kurtlogan: Refactor this file to use typed identifiers instead of strings, this will allow you to write a more generic function instead of creating a new function every time something needs added to the cache. The following would work
```
trait TypedIdentifier[A] {
  val id: String
}

abstract class AddrTypeId[A](partialId: String) extends TypedIdentifier[A] {
  val typ: AddrType
  val id: String = s"$typ$partialId"
}

case class RecordSetId(typ: AddrType) extends AddrTypeId[RecordSet]("SelectedRecordSet")

case object AddressPageVisitedId extends TypedIdentifier[AddressPageVisitedDto] {
  val id: String = "addressPageVisitedDto"
}
```
This would allow you to create a generic function for caching data like so:
```
def addToCache[A](id: TypedIdentifier[A], record: A)(implicit hc: HeaderCarrier): Future[CacheMap] =
  sessionCache.cache(id.id, record)
```

PR #358, opened 2019-11-15
DDCNLS-7089 | Implement Cred strength check for MFA uplift

jpatton638:


conf/application.conf
```diff
@@ -237,6 +237,9 @@ external-url {
   identity-verification {
     host = "http://localhost:9948"
   }
+  bas-gateway-frontend {
+    host = "http://localhost:9553"
+  }
```
oc-ste: has this been added to other configs?

PR #360, opened 2019-11-20
Added a new route to the old ambig journey

oc-ste:


test/controllers/auth/EnforceAmbiguousUserActionSpec.scala
```diff
@@ -55,44 +56,19 @@ class EnforceAmbiguousUserActionSpec extends FreeSpec with MustMatchers with Gui
...
-            None,
-            None,
-            None,
-            None,
-            None,
-            FakeRequest()
-          )
-        val result = harness()(userRequest)
-        status(result) mustBe OK
+        Seq(NotEnrolledSelfAssessmentUser(SaUtr("1111111111")), WrongCredentialsSelfAssessmentUser(SaUtr("1111111111")))
```
kurtlogan: I would be tempted to have the map around the test, as it stands if the test fails it won't give you much information about what has failed but if you wrap the test in the map then `.toString` the value in the name of the test and it will help identify which value it fails for

app/views/selfAssessmentNotShown.scala.html
```diff
@@ -19,7 +19,7 @@
 @import uk.gov.hmrc.renderer.TemplateRenderer
 @import util.LocalPartialRetriever
 
-@(utr: SaUtr)(implicit request: UserRequest[_], configDecorator: ConfigDecorator, partialRetriever: LocalPartialRetriever, templateRenderer: TemplateRenderer, messages: play.api.i18n.Messages)
+@(utr: Option[String])(implicit request: UserRequest[_], configDecorator: ConfigDecorator, partialRetriever: LocalPartialRetriever, templateRenderer: TemplateRenderer, messages: play.api.i18n.Messages)
```
kurtlogan: Consider changing this to use `Option[Utr]` instead of `Option[String]`. It will give a little more type safety

PR #365, opened 2019-11-27
DDCNLS-7107 | Implement Caching of SA user type to stop multiple call…

jpatton638:
…s to Enrolments


app/models/SelfAssessmentUserType.scala
```diff
@@ -16,15 +16,55 @@
...
   def saUtr: SaUtr
+}
+
+object SelfAssessmentUserType {
+  val cacheId = "SelfAssessmentUser"
+
+  implicit val writes = new Writes[SelfAssessmentUserType] {
+    override def writes(o: SelfAssessmentUserType): JsValue = o match {
+      case ActivatedOnlineFilerSelfAssessmentUser(utr) =>
+        Json.obj("_type" -> JsString("ActivatedOnlineFilerSelfAssessmentUser"), "utr" -> JsString(utr.toString))
```
oc-ste: Better to just ask the object for it's name to ensure it's always the same, if someone were to change the class name it would be a one place fix: 

`ActivatedOnlineFilerSelfAssessmentUser.toString`

app/models/SelfAssessmentUserType.scala
```diff
@@ -16,15 +16,55 @@
...
+  }
 }
+
 case class ActivatedOnlineFilerSelfAssessmentUser(saUtr: SaUtr) extends SelfAssessmentUser
 case class NotYetActivatedOnlineFilerSelfAssessmentUser(saUtr: SaUtr) extends SelfAssessmentUser
 case class WrongCredentialsSelfAssessmentUser(saUtr: SaUtr) extends SelfAssessmentUser
 case class NotEnrolledSelfAssessmentUser(saUtr: SaUtr) extends SelfAssessmentUser
 case object NonFilerSelfAssessmentUser extends SelfAssessmentUserType
+
+case class JsonParseException(str: String) extends Exception
```
oc-ste: Never used

test/services/EnrolmentStoreCachingServiceSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.mockito.MockitoSugar
+import play.api.libs.json.Json
+import uk.gov.hmrc.domain.SaUtr
+import uk.gov.hmrc.http.cache.client.CacheMap
+import util.BaseSpec
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
```
oc-ste: You are just testing the mock returns whatever you pass it in these tests

The spec will have more value when you move `enrolledOnDifferentCredentials` into the EnrolmentStoreCachingService

test/controllers/auth/SelfAssessmentStatusActionSpec.scala
```diff
@@ -107,47 +113,88 @@ class SelfAssessmentStatusActionSpec
...
+        "and they have an enrolment on another credential" - {
+          "return WrongCredentialsSelfAssessmentUser" in {
+            implicit val request = createAuthenticatedRequest(None)
+
+            val saUtr = SaUtr("1111111111")
+
+            when(enrolmentsCachingService.getSaUserTypeFromCache()(any(), any()))
+              .thenReturn(Future.successful(None))
+            when(enrolmentsCachingService.addSaUserTypeToCache(any[SelfAssessmentUserType]())(any(), any()))
+              .thenReturn(Future.successful(CacheMap("id", Map.empty)))
```
oc-ste: A lot of mocking going on, when this happens it suggests we're just testing implementation rather than behaviour. 

I'll re-review this bit when the enrolmentsCachingService logic is updated (as mentioned above)

test/services/EnrolmentStoreCachingServiceSpec.scala
```diff
@@ -50,43 +53,39 @@ class EnrolmentStoreCachingServiceSpec extends BaseSpec with MockitoSugar with S
...
+      "return WrongCredentialsSelfAssessmentUser when the connector returns a Right with a non-empty sequence" in new LocalSetup {
 
-  }
+        override val connectorResult: Either[String, Seq[String]] = Right(Seq[String]("Hello there"))
 
+        sut.getSaUserTypeFromCache(saUtr).futureValue shouldBe WrongCredentialsSelfAssessmentUser(saUtr)
+      }
+    }
+  }
 }
```
oc-ste: Missing a test on the cache, basically `verify` call on` enrolmentsConnector.getUserIdsWithEnrolments` to determine when something has actually been cached.

So the test would be named something like "calling same function twice, only calls the cache once"

PR #366, opened 2019-11-28
DDCNLS-7089 | Add MFA link and check cred strength first

jpatton638:


conf/application.conf
```diff
@@ -238,6 +238,9 @@ external-url {
   tracking-frontend {
     host = "http://localhost:9242"
   }
+  bas-gateway-frontend {
```
kurtlogan: Has this been changed in app config base?

app/controllers/auth/AuthAction.scala
```diff
@@ -44,9 +45,11 @@ class AuthActionImpl @Inject()(
 
   override def invokeBlock[A](request: Request[A], block: AuthenticatedRequest[A] => Future[Result]): Future[Result] = {
 
+    val compositePredicate = CompositePredicate(CredentialStrength(CredentialStrength.strong), ConfidenceLevel.L200)
```
kurtlogan: You can use `and` instead of construcing a composite predicate  
```suggestion
    val compositePredicate = CredentialStrength(CredentialStrength.strong) and ConfidenceLevel.L200
```

PR #371, opened 2019-12-09
DDCNLS-7080 Redirect URI added to GG Route

RobertLaverick:


app/views/main.scala.html
```diff
@@ -23,6 +23,7 @@
 @import controllers.auth.requests.UserRequest
 @import controllers.helpers.WelshWarningHelper
 @import util.LocalPartialRetriever
+@import java.net.URLEncoder
```
oc-ste: Don't need this

app/views/main.scala.html
```diff
@@ -136,7 +137,15 @@
...
-                profileUrl = request.profile,
+                profileUrl  = request.profile match {
+                    case Some(a) => {
+                        if (!a.contains("redirect_uri"))
+                            Some(a + "?redirect_uri=" + URLEncoder.encode(routes.HomeController.index().absoluteURL()))
+                        else
+                            Some(a)
+                        }
+                    case None => None
+                },
```
oc-ste: profileUrl = request.profile.map(url => if(url.contains("redirect_uri")) {
url + "?redirect_uri=" + configDecorator.get(pertax-frontend) + (routes.HomeController.index())
else {
url 
}
}

test/controllers/auth/AuthActionSpec.scala
```diff
@@ -266,4 +266,44 @@ class AuthActionSpec extends FreeSpec with MustMatchers with MockitoSugar with O
         s"Some(TrustedHelper(principalName,attorneyName,returnUrl,$fakePrincipalNino))")
     }
   }
+
+  "A user with a SCP Profile Url must include a redirect uri back to the home controller" in {
```
kurtlogan: What would happen if you provided an invalid url?

app/views/main.scala.html
```diff
@@ -136,7 +136,7 @@
                 )),
                 signoutUrl = Some(controllers.routes.ApplicationController.signout(Some(SafeRedirectUrl(configDecorator.getFeedbackSurveyUrl(configDecorator.defaultOrigin))), None).url),
                 activeTab = request.activeTab,
-                profileUrl = request.profile,
+                profileUrl  = request.profile,
```
kurtlogan: Undo this change

test/controllers/auth/AuthActionSpec.scala
RobertLaverick: I've put together a fix for it the URL has a scheme as a way to check for an absolute URL, but I'm 100% sure that the SCP Profile Urls in prod are absolute or relative.

RobertLaverick: I've had this confirmed by #team-scp-si. It's always absolute

kurtlogan: We should assume they can be both

kurtlogan: Thats fine the way it is then

PR #374, opened 2019-12-10
Ddcnls 6973

Muhammad-Ali-Ahmad:


app/views/page_not_found_template.scala.html
```diff
@@ -0,0 +1,37 @@
...
+@import util.LocalPartialRetriever
+
+@()(implicit request: UserRequest[_],
+configDecorator: ConfigDecorator, partialRetriever: LocalPartialRetriever, templateRenderer: TemplateRenderer, messages: play.api.i18n.Messages)
+
+@anchor(text: String, url: String) = {<a href="@url" target="_blank">@text</a>}
+
+@main(title = messages("global.page.not.found.error.title"),
+pageName = Some(messages("label.your_personal_tax_account")),
+backlinkUrl = Some("javascript:window.history.back();")
```
oc-ste: Are we allowed to do this and should the not found page have a back link?

app/error/RendersErrors.scala
```diff
@@ -48,13 +48,15 @@ trait RendersErrors extends Results {
...
       case _           => "InternalServerError500"
     }
-
-    Status(statusCode)(
-      views.html.error(
-        s"global.error.$errorKey.title",
-        Some(s"global.error.$errorKey.heading"),
-        List(s"global.error.$errorKey.message")))
-
+    if (statusCode == NOT_FOUND) {
```
oc-ste: I think the only time this is possible is in LocalErrorHandler, where we call futureError(statusCode) on line 47. 

You just need to check if the status code is NOT_FOUND in there and render the correct view.

PR #375, opened 2019-12-17
Ddcnl 231

jpatton638:


app/models/AddressJourneyTTLModels.scala
```diff
@@ -0,0 +1,81 @@
...
+  val editSoleAddress: String = "EditSoleAddress"
+  val editPrimaryAddress: String = "EditPrimaryAddress"
+  val editCorrespondenceAddress: String = "EditCorrespondenceAddress"
+
+  val addressType = "addressType"
+  val expireAt = "expireAt"
+
+  implicit val writes = new OWrites[EditedAddress] {
+    def writes(model: EditedAddress): JsObject = Json.obj(
+      addressType -> model.getClass.getSimpleName,
```
kurtlogan: this is britle in that the writes refers to the model name and the reads refers to a variable, change it so they both refer to the variable name
```suggestion
      addressType -> model match {
        case EditedSoleAddress(_) => editSoleAddress
        ...
      },
```

app/repositories/EditAddressLockRepository.scala
```diff
@@ -21,62 +21,103 @@ import java.time.{OffsetDateTime, ZoneId, ZoneOffset}
...
+      )
     )
+  }
 
-  private[repositories] def getCore[S](selector: BSONDocument): Future[Option[AddressJourneyTTLModel]] =
-    this.collection.flatMap(_.find(selector, None).one[AddressJourneyTTLModel])
-
-  private[repositories] def insertCore(nino: String, date: OffsetDateTime): Future[WriteResult] =
-    this.collection.flatMap(_.insert(ordered = false).one(AddressJourneyTTLModel(nino, toBSONDateTime(date))))
+  private[repositories] def getCore[S](selector: BSONDocument): Future[List[AddressJourneyTTLModel]] =
```
kurtlogan: Don't need type paramater

app/repositories/EditAddressLockRepository.scala
```diff
@@ -21,62 +21,103 @@ import java.time.{OffsetDateTime, ZoneId, ZoneOffset}
...
+        case e: Exception =>
+          Logger.error(s"Unable to find document: ${e.getMessage}")
+          List[AddressJourneyTTLModel]()
+      }
+
+  private[repositories] def insertCore(record: AddressJourneyTTLModel): Future[WriteResult] = {
+    val index = Index(Seq((record.createIndex, IndexType.Ascending)), unique = false)
+
+    this.collection.flatMap(coll =>
+      coll.indexesManager.ensure(index).flatMap { result =>
```
kurtlogan: You don't need to ensure the index on every insert, once when the application starts is enough, maybe put it in the `setIndex` method

app/repositories/EditAddressLockRepository.scala
```diff
@@ -21,62 +21,103 @@ import java.time.{OffsetDateTime, ZoneId, ZoneOffset}
...
+
+    val getExpiredAt = "editedAddress.expireAt"
+
     getCore(
       BSONDocument(
-        BSONDocument("_id"     -> nino),
-        BSONDocument(EXPIRE_AT -> BSONDocument("$gt" -> toBSONDateTime(OffsetDateTime.now()))))
+        BSONDocument("nino" -> nino),
+        BSONDocument(
+          BSONDocument(getExpiredAt -> BSONDocument("$gt" -> toBSONDateTime(OffsetDateTime.now())))
```
kurtlogan: Do you need an expiry at field? Is this not what the ttl is ensuring?

test/controllers/AddressControllerSpec.scala
```diff
@@ -61,7 +61,7 @@ class AddressControllerSpec extends BaseSpec with MockitoSugar {
   val mockAuthJourney = mock[AuthJourney]
   val mockLocalSessionCache = mock[LocalSessionCache]
   val mockCitizenDetailsService = mock[CitizenDetailsService]
-  val mockCorrespondenceAddressLockRepository = mock[CorrespondenceAddressLockRepository]
+  val mockeditAddressLockRepository = mock[EditAddressLockRepository]
```
kurtlogan: Capitalize the `e`

app/models/AddressJourneyTTLModels.scala
```diff
@@ -0,0 +1,81 @@
...
+        addressType <- (json \ addressType).validate[String]
+        expireAt    <- (json \ expireAt).validate[BSONDateTime]
+      } yield
+        addressType match {
+          case `editSoleAddress`           => EditSoleAddress(expireAt)
+          case `editPrimaryAddress`        => EditPrimaryAddress(expireAt)
+          case `editCorrespondenceAddress` => EditCorrespondenceAddress(expireAt)
+        }
+
+    def read(model: EditedAddress): JsObject = Json.obj(
```
kurtlogan: Is this used?

app/controllers/helpers/PersonalDetailsCardGenerator.scala
```diff
@@ -28,16 +28,25 @@ class PersonalDetailsCardGenerator @Inject()(
...
-  def getPersonalDetailsCards(hasCorrespondenceAddressLock: Boolean)(
+  def getPersonalDetailsCards(changedAddressIndicator: List[AddressJourneyTTLModel])(
     implicit request: UserRequest[_],
     configDecorator: ConfigDecorator,
-    messages: play.api.i18n.Messages): Seq[Html] =
+    messages: play.api.i18n.Messages): Seq[Html] = {
+
+    val optionalEditAddress = changedAddressIndicator.map(y => y.editedAddress)
+
+    val mainAddressChangeIndicator = optionalEditAddress.exists(_.isInstanceOf[EditSoleAddress]) || optionalEditAddress
```
kurtlogan: If you remove the expireAt you can use contains instead of exists

app/models/AddressJourneyTTLModels.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import play.api.libs.json._
+import reactivemongo.bson.BSONDateTime
+import reactivemongo.play.json._
+
+case class AddressJourneyTTLModel(nino: String, editedAddress: EditedAddress)
+
+sealed trait EditedAddress {
+  val expireAt: BSONDateTime
+  def getIndex: String
```
kurtlogan: get index doesn't make to much sense as to what the value is. Maybe something like `addressType` would be better

PR #379, opened 2020-01-06
Ddcnl 213

jpatton638:


jpatton638:
This branch is a copy of DDCNL-231

That branch has been approved

app/controllers/AddressController.scala
```diff
@@ -731,7 +732,7 @@ class AddressController @Inject()(
                              "closure_of_correspondence",
                              auditForClosingPostalAddress(closingAddress, personDetails.etag, "correspondence")))
                      _        <- clearCache() //This clears ENTIRE session cache, no way to target individual keys
```
RobertLaverick: Is it worth adding a ticket to enable a more focused clearing of the cache?

jpatton638: Yeah, can we talk to Andy about raising a ticket for this?

PR #398, opened 2020-02-10
Ddcnl 210

oc-ste:


test/controllers/auth/SessionAuditorSpec.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  val testRequest = FakeRequest()
+
+  def authenticatedRequest[A](request: Request[A]) = AuthenticatedRequest[A](
+    Some(Fixtures.fakeNino),
+    None,
+    Credentials("foo", "bar"),
+    ConfidenceLevel.L200,
+    None,
+    Some(new DateTime("2015-10-21T07:28")),
```
RobertLaverick: Is there a reason we're using a precise DateTime here?

PR #400, opened 2020-02-13
[DC-2243] Redirect to preference-frontend to manage paperless.

lxol:


app/controllers/PaperlessPreferencesController.scala
```diff
@@ -51,13 +51,9 @@ class PaperlessPreferencesController @Inject()(
...
-          for {
-            managePrefsPartial <- preferencesFrontendPartialService.getManagePreferencesPartial(
-                                   configDecorator.pertaxFrontendHomeUrl,
-                                   Messages("label.back_to_account_home"))
-          } yield {
-            Ok(views.html.preferences.managePrefs(managePrefsPartial.successfulContentOrEmpty))
-          }
+          Future.successful(
+            Redirect(preferencesFrontendService
+              .getManagePreferencesUrl(configDecorator.pertaxFrontendHomeUrl, Messages("label.back_to_account_home"))))
```
RobertLaverick: I'm not sure why we need this service to simply return a URL. If we're going to have a simple redirect then why not simply build the url here and test that we've redirected to the expected url. We don't really need a service to build a redirect url imo.

I'd recommend that if you're happy with renaming the service, to remove it since it isn't a service that returns a Partial (and it a bit out of place in a partials package). Then just change the test to confirm the redirect url rather than a call to this service.

lxol: I removed partials.PreferncessFrontendService and updated PR

app/controllers/PaperlessPreferencesController.scala
```diff
@@ -16,28 +16,36 @@
...
   withBreadcrumbAction: WithBreadcrumbAction)(
   implicit partialRetriever: LocalPartialRetriever,
   configDecorator: ConfigDecorator,
   templateRenderer: TemplateRenderer)
-    extends PertaxBaseController {
+    extends PertaxBaseController with ServicesConfig {
+
+  val mode: Mode = environment.mode
+  val runModeConfiguration: Configuration = configuration
+  val preferencesFrontendUrl = baseUrl("preferences-frontend")
```
oc-ste: Can this be moved to ConfigDecorator? It would mean you don't have to extend with ServicesConfig

test/controllers/PaperlessPreferencesControllerSpec.scala
```diff
@@ -74,8 +63,7 @@ class PaperlessPreferencesControllerSpec extends BaseSpec with MockitoSugar {
         })
 
         val r = controller.managePreferences(FakeRequest())
-        status(r) shouldBe OK
-        verify(controller.preferencesFrontendPartialService, times(1)).getManagePreferencesPartial(any(), any())(any())
+        status(r) shouldBe SEE_OTHER
```
oc-ste: Assert the redirect URL, either by mocking Tools or asserting the start of the Url

oc-ste:
Note that there is a merge conflict on this branch

app/controllers/PaperlessPreferencesController.scala
lxol: done

lxol: done

lxol:
I resolved the conflict and updated PR

PR #401, opened 2020-03-03
DDNCL-1582 | Handle users with Cred Strength 'none' correctly

jpatton638:


app/controllers/auth/AuthAction.scala
```diff
@@ -130,7 +129,7 @@ class AuthActionImpl @Inject()(
         case _ => throw new RuntimeException("Can't find credentials for user")
       }
   } recover {
-    case _: NoActiveSession =>
+    case _: NoActiveSession | _: IncorrectCredentialStrength =>
```
oc-ste: Having another look at this, I think all we want to do for this exception is: Redirect(configDecorator.authProviderChoice)

app/views/main.scala.html
```diff
@@ -119,10 +119,6 @@
     }
 
     @{
-
-        val prev = LanguageHelper.langUtils.Dates.formatEasyReadingTimestamp(request.previousLoginTime, "")
```
oc-ste: Remove logintime from auth as well

PR #410, opened 2020-04-06
Ddcnl 1725

jpatton638:
This is to merge onto the Play-2.6 branch after that has gone in.
It will be easier reviewing commit by commit, as each commit covers the migration of one controller with tests.


test/controllers/address/StartDateControllerSpec.scala
```diff
@@ -0,0 +1,324 @@
...
+
+      val result = controller.onPageLoad(PrimaryAddrType)(FakeRequest())
+
+      status(result) shouldBe SEE_OTHER
+      redirectLocation(result) shouldBe Some("/personal-account/personal-details")
+      verify(controller.sessionCache, times(1)).fetch()(any(), any())
+    }
+  }
+
+  "Calling AddressController.processEnterStartDate" should {
```
oc-ste, line 122: "onSubmit" should { ...

PR #448, opened 2020-06-10
Ddcnl 1984 Index out of bounds for Address

AndrewArkless:


app/models/addresslookup/AddressRecord.scala
```diff
@@ -23,6 +23,8 @@ import play.api.libs.json.Json
   */
 case class AddressRecord(id: String, address: Address, language: String) {
 
+  def validAddress = address.hasAddressLines
+
   def isValid: Boolean = address.isValid && language.length == 2
```
oc-ste, line 28: Is there a way to combine these two? It could get confusing to use AddressRecord as it's now ambiguous as to what a valid address is

app/models/addresslookup/Address.scala
```diff
@@ -31,6 +31,8 @@ case class Address(
   country: Country,
   subdivision: Option[String]) {
 
+  def hasAddressLines: Boolean = lines.nonEmpty
+
   def isValid: Boolean = lines.nonEmpty && lines.size <= (if (town.isEmpty) 4 else 3)
```
oc-ste, line 36: refactor lines.nonEmpty to be hasAddressLines

app/models/addresslookup/AddressRecord.scala
AndrewArkless: It is, I was thinking of another ticket for this, the old code, it is not being used anywhere and it has validation logic beyond the remit of the ticket of which we don't know if it is right or the reasons it's not used. Agree it is confusing and I would like it to come out.

PR #453, opened 2020-06-29
Ddcnl 1826

jpatton638:


app/controllers/ApplicationController.scala
```diff
@@ -44,7 +45,9 @@ class ApplicationController @Inject()(
   failedIvIncompleteView: FailedIvIncompleteView,
   lockedOutView: LockedOutView,
   timeOutView: TimeOutView,
-  technicalIssuesView: TechnicalIssuesView)(
+  technicalIssuesView: TechnicalIssuesView,
+  val notFoundView: NotFoundView,
+  val errorView: ErrorView)(
```
oc-ste, line 50: I think the better fix is to remove the RenderErrors trait, turn it into a class and DI it in wherever these two views (or RenderError functions) are required

app/controllers/address/AddressController.scala
```diff
@@ -64,7 +65,7 @@ abstract class AddressController @Inject()(
 
   def internalServerError(implicit userRequest: UserRequest[_]): Result =
```
oc-ste, line 66: Move this to the RenderErrors thing and it will mean you won't have to inject in ErrorViews here and will tidy things up in general

test/views/html/MainViewSpec.scala
```diff
@@ -0,0 +1,217 @@
...
+import uk.gov.hmrc.renderer.TemplateRenderer
+import util.Fixtures.{buildFakeRequestWithVerify, fakeNino}
+import util.UserRequestFixture.buildUserRequest
+
+class MainViewSpec extends ViewSpec {
+
+  override protected def localGuiceApplicationBuilder(): GuiceApplicationBuilder =
+    GuiceApplicationBuilder()
+      .overrides(
+        bind(classOf[TemplateRenderer]).to(classOf[LocalTemplateRenderer])
```
oc-ste, line 40: LocalTemplateRenderer requires the service to be running, best move these tests to the IT folder

PR #454, opened 2020-07-02
DDCNL-2225: Report partial load failure as errors

oc-ste:


oc-ste:
@pascal22p this is what the logging will look like for load partial failures

PR #472, opened 2020-09-14
Ddcnl 2062 New page for Postal International Address Choice

AndrewArkless:


test/views/html/address/PostalInternationalAddressChoiceViewSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+  "rendering PostalInternationalAddressChoiceView" should {
+    "must render the correct h1 appropriate to postal address" in {
+      val result = asDocument(view(InternationalAddressChoiceDto.form).toString)
+      assertContainsText(result, messages("label.is_your_postal_address_in_the_uk"))
+      assertNotContainText(result, messages("label.do_you_live_in_the_uk"))
+    }
+
+    "must not render the h1 appropriate to residential addresses" in {
+      val result = asDocument(view(InternationalAddressChoiceDto.form).toString)
+      assertNotContainText(result, messages("label.do_you_live_in_the_uk"))
```
oc-ste, line 47: Not assertions tend to miss the point of the test, can you also assert for the presence of something as well?

AndrewArkless: The presence of what should be there is asserted in the previous test. They could be combined, or I could get rid of the Not assertion. My thinking was this enforced the distinction between Postal and Non Postal, that the Non Postal text must not appear on the Postal.

AndrewArkless: ooh one minute

AndrewArkless: I see what I've done I should have taken the Not Assert out in the previous test. So two test one what should be the other what shouldn't. Depends if we like two asserts in one test, or keep each assert seperate.

oc-ste: I think I'd leave them separate

oc-ste (approved):

PR #481, opened 2020-11-13
DDCNL-2330 | Implement ATS tile on PTA

jpatton638:


jpatton638:
Still waiting for the Welsh translation for the tile, due for COB 16/11

test/controllers/controllershelpers/HomeCardGeneratorSpec.scala
```diff
@@ -52,10 +58,17 @@ class HomeCardGeneratorSpec extends ViewSpec with MockitoSugar {
...
       marriageAllowance,
-      statePension
+      statePension,
+      taxSummaries
     )
   val testUtr = SaUtr(new SaUtrGenerator().nextSaUtr.utr)
 
+  override implicit lazy val app: Application = new GuiceApplicationBuilder()
+    .configure(
+      "feature.tax-summaries-tile.enabled" -> true
```
struong, line 68: Don't need this block as it's true by default

app/controllers/controllershelpers/HomeCardGenerator.scala
```diff
@@ -50,7 +51,8 @@ class HomeCardGenerator @Inject()(
       getTaxCalculationCard(taxCalculationStateCyMinusOne),
       getTaxCalculationCard(taxCalculationStateCyMinusTwo),
       getSelfAssessmentCard(saActionNeeded, currentTaxYear + 1),
-      getNationalInsuranceCard()
+      getNationalInsuranceCard(),
+      getAnnualTaxSummaryCard
```
struong, line 55: Has this got tests around it? (fine if not as its a bit of an odd function)

jpatton638: Each of the functions called here are tested, and this just returns the result of these functions, so it should be fine

struong (approved):

shikkhasingh (approved):

shikkhasingh (approved):

shikkhasingh (approved):

shikkhasingh (approved):

PR #503, opened 2021-03-15
DDCNL-3689

brendanstraughan:


app/controllers/auth/AuthAction.scala
```diff
@@ -51,14 +51,14 @@ class AuthActionImpl @Inject()(
       res.replaceParams("redirect_uri", configDecorator.pertaxFrontendBackLink).toString()
     }
 
-  object GT100 {
+  object LT200 {
```
struong, line 54: Need to update the spec L100 no longer exists so needs to be removed from the code base and changed to 50 or 200 dependent on the scenario

struong (approved):

PR #557, opened 2022-02-25
user research banner added to childbenefit page

Muhammad-Ali-Ahmad:


app/assets/javascripts/pertax.js
```diff
@@ -1,9 +1,9 @@
 /* global $, GOVUK */
-$('.full-width-banner__close, .covid-banner__close').click(function () {
+$('.covid-banner__close').click(function () {
   $.ajax({
     url: '/personal-account/dismiss-ur-banner',
```
Muhammad-Ali-Ahmad, line 4: I looked in to using this but because it is same function, if a user clicks hide covid banner, the variable to display banner is set to false and UR banner won't be displayed. Likewise if a user clicks hide UR banner, the covid banner will also be hidden.
For this reason I have set a cookie "mdtpurr" for UR banner as we did in the past that lasts for 1 month as per the guidance https://www.staging.tax.service.gov.uk/help/cookie-details

pascal22p (approved):

PR #569, opened 2022-03-24
Ddcnl 5409

Sachin300793:


app/config/ConfigDecorator.scala
```diff
@@ -267,6 +267,10 @@ class ConfigDecorator @Inject() (
   lazy val sessionTimeoutInMinutes = sessionTimeoutInSeconds / 60
   lazy val sessionCountdownInSeconds = runModeConfiguration.getOptional[Int]("ptaSession.countdown").getOrElse(120)
 
+  lazy val incomeTaxViewChangeFrtonendHost = getExternalUrl(s"income-tax-view-change-frontend.host").getOrElse("")
+
+  lazy val itsaViewUrl = s"$incomeTaxViewChangeFrtonendHost/report-quarterly/income-and-expenses/view"
```
pascal22p, line 272: The origin is missing on the itsa url:
/report-quarterly/income-and-expenses/view?origin=PTA

conf/messages
```diff
@@ -448,7 +448,7 @@ label.make_a_payment=Make a payment
 label.view_your_payments=View your payments
 label.check_if_you_need_to_fill_in_a_tax_return=Check if you need to fill in a tax return
 label.find_out_how_to_access_self_assessment=Find out how to access Self Assessment
-label.view_and_manage_your_self_assessment_tax_return_the_deadline_for_online_=View and manage your Self Assessment tax return. The deadline for online returns is 31 January {0}.
+label.view_and_manage_your_self_assessment_tax_return_the_deadline_for_online_=View and manage your Self Assessment tax return.
```
pascal22p, line 451: The old label should be kept for the old sa tile.

pascal22p (changes requested):

conf/messages
SandeepUmredkar: We have broken down the label in two parts. The label still remains intact. But first part being used in multiple places hence we did this. I think we forgot to update the old SA tile to reflect using 2 keys. will sort it out.

app/config/ConfigDecorator.scala
Sachin300793: Done

PR #571, opened 2022-03-25
DDCNL-5409: new ITSA tile

pascal22p:


test/controllers/auth/GetPersonDetailsActionSpec.scala
```diff
@@ -60,6 +60,7 @@ class GetPersonDetailsActionSpec extends BaseSpec {
       ConfidenceLevel.L50,
       None,
       None,
+      null,
```
Christian-QA, line 63: What argument is this? Surprised you can have a null here

Christian-QA (approved):

test/controllers/auth/GetPersonDetailsActionSpec.scala
Sachin300793: Replaced null with Set() - empty set , thanks

Christian-QA (approved):

Christian-QA (approved):

Luka-Abey (approved):

PR #592, opened 2022-05-09
Update banner

pascal22p:


conf/application.conf
```diff
@@ -352,6 +352,17 @@ feature {
     maxTps = 10
     timeoutInSec = 5
   }
+
+  banner {
+    home {
+      enabled = true
+      heading.en = "Help improve GOV.UK"
```
Sachin300793, line 359: Can we put heading text in Messages ?

Sachin300793, line 359: Can we put heading text in Message or any specific reason ?

pascal22p: If you put the text in message then it is part of the codebase which I don't want. The text should be configurable at runtime so via app-config-*.
Which means no code change, no  lengthy QA check to push a new banner except a design check is staging. It speeds up massively the change of the banner.

Sachin300793: understood , great

Sachin300793 (approved):

Luka-Abey (approved):

PR #610, opened 2022-06-20
DDCNL-5574: Removing the config from application.conf file for news a…

SandeepUmredkar:
DDCNL-5574: Removing the config from application.conf file for news and changing the code for handling of null. And removing of test case as the default values are not there so we cannot test that part in the view

### DIFF
--- a/app/config/NewsAndTilesConfig.scala
+++ b/app/config/NewsAndTilesConfig.scala
@@ -27,43 +27,51 @@ import java.time.format.DateTimeFormatter
 import scala.collection.JavaConverters._
 
 @Singleton
-class NewsAndTilesConfig @Inject() (configuration: Configuration, localDateUtilities: LocalDateUtilities) {
+class NewsAndTilesConfig @Inject()(configuration: Configuration, localDateUtilities: LocalDateUtilities) {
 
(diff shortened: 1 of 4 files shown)


app/config/NewsAndTilesConfig.scala
```diff
@@ -27,43 +27,51 @@ import java.time.format.DateTimeFormatter
...
-        } else {
-          None
         }
-      }
-      .toList
-      .flatten
+        .toList
+        .flatten
+    } else {
+      List[NewsAndContentModel]()
```
pascal22p, line 73:
```suggestion
      List.empty
```

test/views/html/cards/LatestNewsAndUpdatesViewSpec.scala
```diff
@@ -44,20 +44,5 @@ class LatestNewsAndUpdatesViewSpec extends ViewSpec {
         Messages("label.latest_news_and_updates")
       )
     }
-
```
pascal22p, line 47: Keep the tests. Just override the configValues in the localGuiceApplicationBuilder

pascal22p (changes requested):

pascal22p (approved):

PR #611, opened 2022-06-23
DDCNL-6091 :  Adding of mock for HomePageCachingHelper so it does not call LocalSessionCaching

SandeepUmredkar:
… call LocalSeccionCache


Sachin300793 (approved):

SandeepUmredkar:
Merging it directly to main, as this is just change in UTs and does not need to be QAed

PR #644, opened 2022-10-07
autocomplete added

Muhammad-Ali-Ahmad:


app/views/personaldetails/UpdateInternationalAddressView.scala.html
```diff
@@ -157,5 +151,9 @@
         label = Label(
             content = Text(messages("label.country"))
         )
-    ))
+    ).asAccessibleAutocomplete(Some(
+        AccessibleAutocomplete(
+            showAllValues = true,
+            autoSelect = false)
```
oscarduignan, line 157: @Muhammad-Ali-Ahmad you've got the error I slipped in here, where I reversed the true / false values from what you had previously https://github.com/hmrc/pertax-frontend/pull/644/files#diff-981f9d2ccdce61c6dd47a0f85883664b2937dd550c27a3ced1c7ee9096aff111L52-L53 I think that should make the behaviour match what you had before - I don't know what's causing the width to extend out, doesn't look like you had any custom css in https://github.com/hmrc/pertax-frontend/blob/main/app/assets/stylesheets/modules/accessible-autocomplete.min.scss I'll check in with you next week to see if it's a bug if that's ok 👍

Muhammad-Ali-Ahmad: done

Christian-QA (dismissed):

pascal22p (approved):

PR #671, opened 2022-12-14
DDCNL: deleted custom metrics, clean up code

pascal22p:


project/AppDependencies.scala
```diff
@@ -9,7 +9,7 @@ object AppDependencies {
 
   val compile: Seq[ModuleID] = Seq(
     ws,
-    "uk.gov.hmrc"       %% s"bootstrap-frontend-$playVersion" % "5.20.0",
+    "uk.gov.hmrc"       %% s"bootstrap-frontend-$playVersion" % "6.3.0",
```
Christian-QA, line 12: Can bootstrap go up to 7.12.0 in this ticket? Or is there a butterfly effect beyond scope?

Christian-QA (dismissed):

project/AppDependencies.scala
pascal22p: not in this ticket. that change should not be there it slept through from another ticket

PR #698, opened 2023-03-13
DDCNL-7110 : Fixed exception due to trusted helper from taxcalc

askar-makkunnath:


test/controllers/HomeControllerSpec.scala
```diff
@@ -88,6 +89,7 @@ class HomeControllerSpec extends BaseSpec with CurrentTaxYear {
     lazy val confidenceLevel: ConfidenceLevel     = ConfidenceLevel.L200
     lazy val withPaye: Boolean                    = true
     lazy val year                                 = 2017
+    lazy val trustedHelper: Option[TrustedHelper] = None
```
pascal22p, line 92: All the tests are using None. A test is missing when trusted helper is not None

askar-makkunnath: Added spec for the scenario when a trusted helper is in use.

pascal22p (dismissed):

PR #702, opened 2023-04-11
DDCNL-7027

SandeepUmredkar:


pascal22p (changes requested):
Banner content should be in app-config-* so we don't need to build a new release

PR #717, opened 2023-06-12
DDCNL-7468

SandeepUmredkar:


conf/application.conf
```diff
@@ -199,7 +199,10 @@ microservice {
       host = localhost
       port = 9238
     }
-
+    save-your-national-insurance-number {
+      host = localhost
+      port = 9019
+    }
```
pascal22p, line 205: there is no http call from pertax-frontend to save-your-national-insurance-number microservice.

This is used to create an html link and should be in external-url section

app/controllers/NiLetterController.scala
```diff
@@ -39,24 +41,31 @@ class NiLetterController @Inject() (
   errorRenderer: ErrorRenderer,
   printNiNumberView: PrintNationalInsuranceNumberView,
   pdfWrapperView: NiLetterPDfWrapperView,
-  niLetterView: NiLetterView
+  niLetterView: NiLetterView,
+  featureFlagService: FeatureFlagService
 )(implicit configDecorator: ConfigDecorator, val ec: ExecutionContext)
     extends PertaxBaseController(cc) {
 
   def printNationalInsuranceNumber: Action[AnyContent] =
```
pascal22p, line 49: I would not have the change in here but there:
https://github.com/hmrc/pertax-frontend/blob/6b2065329cce196e30b549f779a19e6655653346/app/viewmodels/PersonalDetailsViewModel.scala#L159

pascal22p (changes requested):

Christian-QA (approved):

pascal22p (approved):

PR #723, opened 2023-06-26
DDCNL-7495

SandeepUmredkar:


app/config/ConfigDecorator.scala
```diff
@@ -315,12 +315,18 @@ class ConfigDecorator @Inject() (
...
-  lazy val breathingSpcaeBaseUrl: String       = servicesConfig.baseUrl("breathing-space-if-proxy")
-  lazy val breathingSpaceTimeoutInSec: Int     =
+  lazy val breathingSpcaeBaseUrl: String        = servicesConfig.baseUrl("breathing-space-if-proxy")
+  lazy val breathingSpaceTimeoutInSec: Int      =
     servicesConfig.getInt("feature.breathing-space-indicator.timeoutInSec")
-  lazy val preferenceFrontendTimeoutInSec: Int =
+  lazy val preferenceFrontendTimeoutInSec: Int  =
     servicesConfig.getInt("feature.preferences-frontend.timeoutInSec")
-  lazy val ptaNinoSaveUrl                      = "https://www.tax.service.gov.uk/save-your-national-insurance-number"
+  lazy val ptaNinoSaveUrl                       = "https://www.tax.service.gov.uk/save-your-national-insurance-number"
```
pascal22p, line 323: Do not use an absolute url for tax.service.gov.uk it does not work locally.
Use the external-url section where the host is overriden in app-config-base.

pascal22p (changes requested):

app/config/ConfigDecorator.scala
SandeepUmredkar: @pascal22p , this change is not part of this ticket. This has already been taken care as part of 7468, which is in testing at the moment

SandeepUmredkar: It might be showing in this PR because of alignment change or something

pascal22p (approved):

PR #731, opened 2023-08-02
Added The Republic of North Macedonia to the list of countries

pascal22p:


pascal22p:
Country list needs to match the one in citizen-details which turn it into a country code.

PR #780, opened 2023-12-13
Ddcnl 8107

grantrobson:


it/controllers/TimeoutsISpec.scala
```diff
@@ -0,0 +1,321 @@
...
+    when(mockFeatureFlagService.get(ArgumentMatchers.eq(TaxComponentsToggle)))
+      .thenReturn(Future.successful(FeatureFlag(TaxComponentsToggle, isEnabled = true)))
+  }
+
+  "/personal-account" must {
+    "hide breathing space related components when breathing space connector times out" in {
+      val result            = getHomePageWithAllTimeouts
+      val content: Document = Jsoup.parse(contentAsString(result))
+      content.getElementsByClass("hmrc-caption govuk-caption-xl").get(0).text() mustBe
+        Messages("label.this.section.is") + " " + Messages("label.account_home")
```
Nj147, line 147: Could you swap out the messages calls in this file with the expected text? Makes things bit more reliable

Nj147 (changes requested):

it/controllers/TimeoutsISpec.scala
grantrobson: OK, done

Nj147 (approved):

PR #802, opened 2024-04-02
DDCNL-8690: refactor news content

SandeepUmredkar:


conf/application.conf
```diff
@@ -332,6 +332,9 @@ feature {
   address-lookup {
       timeoutInSec = 5
   }
+  override-start-and-end-dates-for-news-items {
+    enabled = true
```
pascal22p, line 336: Default should be false so that no override is needed in production

pascal22p (changes requested):

Nj147 (approved):

pascal22p (approved):

PR #806, opened 2024-04-18
Ddcnl 8732

SandeepUmredkar:


conf/application.conf
```diff
@@ -339,8 +319,6 @@ feature {
       enabled = false
       heading.en = "Help make GOV.UK better"
       link.text.en = "Sign up to take part in research (opens in new tab)"
```
pascal22p, line 321: Weird. The welsh has been deleted but not the english?

pascal22p (changes requested):

conf/application.conf
SandeepUmredkar: I believe this is not supposed to be removed and is removed by error. I will add it back,

pascal22p (approved):

PR #827, opened 2024-06-28
Ddcnl 9033: PTA - Rewrite HomeControllerSpec

SandeepUmredkar:


app/config/Filters.scala
```diff
@@ -0,0 +1,47 @@
+/*
```
pascal22p, line 1: This file should not be here.

conf/application.conf
```diff
@@ -40,7 +40,7 @@ sosOrigin = "PERTAX"
 
 http-verbs.retries.ssl-engine-closed-already.enabled = true
 
-play.filters.enabled += "uk.gov.hmrc.sca.filters.WrapperDataFilter"
+play.http.filters = "config.Filters"
```
pascal22p, line 43: This should be as in main `uk.gov.hmrc.sca.filters.WrapperDataFilter`

pascal22p (changes requested):

pascal22p:
This PR is empty

PR #829, opened 2024-07-03
DDCNL-8316: Removed Tax Credits from outage list and re-arranged to alphabetical order

dannybassi:
…order


mathivanan-kailasam (approved):

conf/messages
```diff
@@ -686,8 +686,6 @@ alertBannerShuttering.repayments = Repayments
 alertBannerShuttering.taxcalc = Tax Calculation
 alertBannerShuttering.fandf = Trusted Helpers
 alertBannerShuttering.childBenefit = Child Benefit
```
SandeepUmredkar, line 688: Do we need child benefit property, as you removed that as well from shuttering view

dannybassi: Child benefit is still there, it just moved higher up so its in alphabetical order

PR #830, opened 2024-07-12
DDCNL-8376

SandeepUmredkar:


test/connectors/AddressLookupConnectorSpec.scala
```diff
@@ -157,5 +168,40 @@ class AddressLookupConnectorSpec extends BaseSpec with WireMockHelper with Integ
         result.value.futureValue.swap.getOrElse(UpstreamErrorResponse("", OK)).statusCode mustBe httpResponse
       }
     }
+
```
grantrobson, line 171: tbh I think all we need for all of these files is one test case for the left, one for the right and one for future failed. All the different response codes should be tested in HttpClientResponseSpec - and if not just add them to the latter

SandeepUmredkar: Ok. The impression I got after reading the ticket is that we have to get rid of wiremocking if they can be easily be mocked without wiremocking. Hence I separated the conditions tested in HttpClientResponseSpec and the rest of it. Will discuss with you tomorrow on what should be done.

grantrobson: I'm ok with what you've done Sandeep. I'll approve the PR.

grantrobson (approved):

PR #848, opened 2024-10-22
DDCNL-9561

SandeepUmredkar:


app/controllers/controllershelpers/HomeCardGenerator.scala
```diff
@@ -95,23 +96,24 @@ class HomeCardGenerator @Inject() (
...
-        )
-      )
-    } else {
-      None
+  ): Option[HtmlFormat.Appendable] = {
+    val isItsaEnrolled = enrolmentsHelper.itsaEnrolmentStatus(request.enrolments).isDefined
+    val saView         = saAndItsaMergeView(
+      (current.currentYear + 1).toString,
+      isItsaEnrolled
+    )
```
grantrobson, line 107: These could do with making a bit clearer. Maybe using if might be better than match? Also method name a bit long

app/controllers/controllershelpers/HomeCardGenerator.scala
```diff
@@ -96,22 +96,25 @@ class HomeCardGenerator @Inject() (
...
-    )
+    val saView = saAndItsaMergeView((current.currentYear + 1).toString, isItsaEnrolled)
 
-    request.trustedHelper.isEmpty match {
-      case false                                  => None
-      case true if isItsaEnrolled || request.isSa => Some(saView)
-      case true if configDecorator.pegaEnabled    =>
-        Some(selfAssessmentRegistrationView()) // Temporary pegaEnabled condition
-      case true                                   => None // In the future, this will be removed when pegaEnabled condition is removed
+    if (request.trustedHelper.isDefined) {
```
grantrobson, line 107: You could just map/ flatMap this. A None => None

SandeepUmredkar: Are you suggesting to use fold for request.trustedHelper, or request.trustedHelper.flatMap(_ => None).orElse {
      if (isItsaEnrolled || request.isSa) {
        Some(saView)
      } else if (configDecorator.pegaEnabled) {
        // Temporary condition for Pega
        Some(selfAssessmentRegistrationView())
      } else {
        None
      }
    }

I wanted to keep all logic in one type like if elses
, hence did that. but can do what you are suugesting

grantrobson: request.trustedHelper.flatMap{ _ =>
      if (isItsaEnrolled || request.isSa) {
        Some(saView)
      } else if (configDecorator.pegaEnabled) {
        // Temporary condition for Pega
        Some(selfAssessmentRegistrationView())
      } else {
        None
      }
    }

is one line less. However, I'm OK with the way you've done it too.

conf/application.conf
```diff
@@ -315,6 +315,9 @@ feature {
   override-start-and-end-dates-for-news-items {
     enabled = false
   }
+  pega{
+   enabled = false
```
grantrobson, line 319: Maybe rename to interstitialPageEnabled - since I think we will have further pega features in future

grantrobson (dismissed):

app/controllers/controllershelpers/HomeCardGenerator.scala
SandeepUmredkar: isn't your code the opposite of what we would expect. It would go on to check for other things even if the trustedhelper is defined while we need it to return None

conf/application.conf
SandeepUmredkar: Good idea

app/controllers/controllershelpers/HomeCardGenerator.scala
grantrobson: Just stick with what you got Sandeep, it's fine

grantrobson (approved):

app/controllers/controllershelpers/HomeCardGenerator.scala
```diff
@@ -104,7 +104,7 @@ class HomeCardGenerator @Inject() (
     val isItsaEnrolled = enrolmentsHelper.itsaEnrolmentStatus(request.enrolments).isDefined
     val saView         = saAndItsaMergeView((current.currentYear + 1).toString, isItsaEnrolled)
 
-    request.trustedHelper.flatMap(_ => None).orElse {
+    request.trustedHelper.map(_ => None).getOrElse {
       if (isItsaEnrolled || request.isSa) {
         Some(saView)
```
grantrobson, line 109: tbh I think this would be clearer if you just say request.trustedHelper match {case Some(_) => None case _ => etc... }

grantrobson: Or you could match like this: (request.trustedHelper, isItsaEnrolled || request.isSa, configDecorator.pegaSaRegistrationEnabled) match

grantrobson (dismissed):

grantrobson (approved):

PR #852, opened 2024-11-12
Ddcnl 9519

grantrobson:


app/controllers/controllershelpers/HomeCardGenerator.scala
```diff
@@ -96,19 +100,40 @@ class HomeCardGenerator @Inject() (
...
 
+  private def displaySACall: Call       = routes.InterstitialController.displaySelfAssessment
+  private def handleSACall: Call        = routes.SelfAssessmentController.handleSelfAssessment
+  private def redirectToEnrolCall: Call = routes.SelfAssessmentController.redirectToEnrolForSa
+
+  private def callAndContent(implicit request: UserRequest[_]) =
+    request.saUserType match {
+      case ActivatedOnlineFilerSelfAssessmentUser(_)       => Some(Tuple2(displaySACall, "label.viewAndManageSA"))
+      case NotYetActivatedOnlineFilerSelfAssessmentUser(_) =>
+        Some(Tuple2(handleSACall, "label.activate_your_self_assessment"))
```
SandeepUmredkar, line 111: I think we can safely remove Tuple2 in all places. Directly using (displaySACall, "label.viewAndManageSA") and so on will be more natural way to write this,

pascal22p (approved):

app/controllers/controllershelpers/HomeCardGenerator.scala
grantrobson: I pulled that logic out of the getSelfAssessmentCard method as there's already quite a bit of conditional logic in there + it was getting hard to follow. I feel a tuple2 is the best way to pass the info back here without duplicating the case logic or adding in complicated functional stuff. But if you want to suggest another way please feel free :-)

dannybassi (approved):

PR #861, opened 2024-12-10
Ddcnl 9719

SandeepUmredkar:


app/config/ConfigDecorator.scala
```diff
@@ -302,4 +303,16 @@ class ConfigDecorator @Inject() (
   lazy val pegaSaRegistrationEnabled: Boolean =
     servicesConfig.getBoolean("feature.pegaSaRegistration.enabled")
 
+  def featureBannerTcsServiceClosure: BannerTcsServiceClosure =
```
grantrobson, line 306: Shouldn't we default this to something rather than throwing an exception?

SandeepUmredkar: Well rather than defaulting and having the wrong state. I thought of throwing en exception would be better. Atleast we would be aware of it. As happened for the feature flag scenario in the past

grantrobson: OK, well in that case don't we need to ensure we have something in app config base/ production etc otherwise won't be it throw exception when we deploy it?

SandeepUmredkar: Yeah, I will make PR for app-config-production if you and/or Pascal is good with the approach. Mainly because of three states. In past somewhere I have seen using 2 boolean flags used for having 3 states. I felt thats not correct. So did it this way.

grantrobson: OK

grantrobson (approved):

PR #900, opened 2025-04-09
DDCNL-10231 Moved trusted helper retrieval to call fandf

Nj147:


test/connectors/FandFConnectorSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+    "return as Some(trustedHelper) when trustedHelper json returned" in new SpecSetup {
+      stubGet("/delegation/get", OK, Some(fandfTrustedHelperResponse))
+
+      val result: Option[TrustedHelper] =
+        connector.getTrustedHelper().value.futureValue.getOrElse(None)
+
+      result mustBe Some(trustedHelper)
+    }
+
+    List(
```
grantrobson, line 72: Can we have a test for fandf endpoint returning NotFound (404)?

app/services/FandfService.scala
```diff
@@ -0,0 +1,37 @@
...
+
+class FandfService @Inject() (
+  fandFConnector: FandFConnector
+) extends Logging {
+
+  def getTrustedHelper()(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[Option[TrustedHelper]] =
+    fandFConnector.getTrustedHelper().foldF(_ => Future.successful(None), helper => Future.successful(helper))
```
grantrobson, line 35: I'm wondering what would happen if we had an intermittent 500 error from fandf - the trusted helper will vanish but only intermittently - could be confusing. Maybe we need to just throw the exception?

app/controllers/auth/AuthRetrievals.scala
```diff
@@ -57,44 +62,45 @@ class AuthRetrievalsImpl @Inject() (
...
+          Some(CredentialStrength.strong) ~
+          confidenceLevel ~
+          profile =>
+        val finalRequest = for {
+          userAnswers   <- journeyCacheRepository.get(hc)
+          trustedHelper <- getTrustedHelper
+        } yield {
+          val trimmedRequest: Request[A] = request
+            .map {
+              case AnyContentAsFormUrlEncoded(data) =>
```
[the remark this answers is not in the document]
Nj147: It's pre-existing stuff, not actually sure what of the request it's actually trimming

grantrobson (approved):

app/controllers/auth/AuthRetrievals.scala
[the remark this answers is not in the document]
grantrobson: Oh yes you are right, it's in main.

grantrobson (approved):

app/services/FandfService.scala
grantrobson, line 35: Could we put a logger.warn here with the exception? As any 5xxx errors will result in no trusted helper in banner. If getting lots of timeouts we need to know so we can increase slots/ memory

Nj147: Added logging for all errors except 404s

PR #906, opened 2025-06-02
10562 Add published date

grantrobson:


pascal22p (dismissed):

SandeepUmredkar (approved):
Changes are good. But we can optimize the imports of ViewNewsAndUpdatesView as ExecutionContext is not used.

grantrobson:
@SandeepUmredkar 
Remove execution context. Well spotted!

PR #907, opened 2025-06-04
Ddcnl 10579

grantrobson:


app/connectors/TaiConnector.scala
```diff
@@ -33,20 +36,29 @@ class TaiConnector @Inject() (
...
+      if (toggle.isEnabled) {
+        val url = s"$taiUrl/tai/$nino/tax-account/$year/tax-components"
+        httpClientResponse
+          .read(
+            httpClientV2
+              .get(url"$url")
+              .transform(_.withRequestTimeout(configDecorator.taiTimeoutInMilliseconds.milliseconds))
+              .execute[Either[UpstreamErrorResponse, HttpResponse]](readEitherOf(readRaw), ec)
+          )
+          .map(result => result.json.asOpt[A](reads))
```
pascal22p, line 58: don't use asOpt. If it fails to parse the json it does not log anything.
It's better to recover to log the exception and return None

app/controllers/HomeController.scala
```diff
@@ -54,11 +58,20 @@ class HomeController @Inject() (
   private val authenticate: ActionBuilder[UserRequest, AnyContent] =
     authJourney.authWithPersonalDetails
 
+  private def getTaxComponentsOrEmptyList(nino: Nino, year: Int)(implicit
```
SandeepUmredkar, line 61: Using Connectors directly in controller might not be good idea specially, where it is doing some data parsing here. But we can do it later if time urgency is there.

SandeepUmredkar (dismissed):
Left one comment if we can accomodate th parsing in new service. would be better.

app/controllers/HomeController.scala
grantrobson: I know what you mean. I removed the service layer as it only had one method in it + wasn't doing anything other than passing straight over to connector so it seemed unnecessary. The data parsing is actually done in the connector (passed in from controller as a reads)

app/controllers/InterstitialController.scala
```diff
@@ -121,9 +130,21 @@ class InterstitialController @Inject() (
...
+  def displayChildBenefitsSingleAccountView: Action[AnyContent] = authenticate.async { implicit request =>
+    taiConnector
+      .taxComponents(request.authNino, current.currentYear)(readsIsHICBCWithCharge)
+      .map(_.getOrElse(false))
+      .fold(_ => false, identity)
+      .map { isRegisteredForHICBCWithCharge =>
+        Ok(
+          viewChildBenefitsSummarySingleAccountInterstitialView(isRegisteredForHICBCWithCharge)
+        )
+      }
```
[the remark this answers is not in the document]
grantrobson: Well spotted :-) - fixed now

app/connectors/TaiConnector.scala
grantrobson: Good point. I've applied change now

app/controllers/HomeController.scala
pascal22p: No need for an empty service.
If you reinstate the service you will need to keep the connector doing only connector thingy.
So move the reads and possibly toggle in the service.
It a simple connector so I am not worried about it. It's fine as it is.

grantrobson: So just checking: you want me to create service, move toggle into it from connector, and reads from controller?

pascal22p (dismissed):

pascal22p (approved):

PR #910, opened 2025-06-13
DDCNL-10607

SandeepUmredkar:


grantrobson (approved):
When we go live with this we need to make sure we switch on in live any toggles that you have changed the name of, e.g. tax components toggle and which are meant to be on in live

app/models/admin/FeatureFlags.scala
```diff
@@ -21,110 +21,115 @@ import uk.gov.hmrc.mongoFeatureToggles.model.{Environment, FeatureFlagName}
...
 
 case object BreathingSpaceIndicatorToggle extends FeatureFlagName {
   override val name: String                = "breathing-space-indicator-toggle"
   override val description: Option[String] = Some(
-    "Enable/disable calls to Breathing space"
+    "Enable/disable the call to Breathing Space service to determine protected debt period status for the user"
   )
 }
 
-case object GetPersonFromCitizenDetailsToggle extends FeatureFlagName {
```
grantrobson, line 89: This toggle is not currently called. But it probably should be -looks like the one place where citizen details is called is not using this flag, so probably should reinstate this flag and call it: https://github.com/hmrc/pertax-frontend/blob/64733811a131fec001298f8953038911c139c218/app/controllers/auth/SelfAssessmentStatusAction.scala#L59

grantrobson (approved):

grantrobson (approved):

PR #924, opened 2025-07-28
Ddcnl 10878

grantrobson:


SandeepUmredkar:
The changes done looks good. but should we not put the same validation for the messageDetails. At the moment its only for message list.

PR #925, opened 2025-07-31
Ddcnl 10846

Nj147:


app/controllers/SaBppInterstitialPageController.scala
```diff
@@ -0,0 +1,64 @@
...
+  view: SPPInterstitialView,
+  appConfig: ConfigDecorator
+) extends PertaxBaseController(cc)
+    with Logging {
+
+  def onPageLoad(isBta: String): Action[AnyContent] = authJourney.authWithPersonalDetails { implicit request =>
+    val origin: String = isBta match {
+      case "true" => "bta-sa"
+      case _      => "pta-sa"
+    }
```
pascal22p, line 39: if (isBta == "bta-sa") true else false

However the origin is always pta so you can remove it

app/controllers/SaBppInterstitialPageController.scala
```diff
@@ -0,0 +1,64 @@
...
+            case Some(SelectSABPPPaymentFormProvider.saBppOverduePayment) =>
+              Redirect(
+                s"${appConfig.bppSpreadTheCostOverduePaymentUrl}?origin=$origin"
+              )
+            case Some(SelectSABPPPaymentFormProvider.saBppAdvancePayment) =>
+              Redirect(
+                s"${appConfig.bppSpreadTheCostAdvancePaymentUrl}?origin=$origin&lang=${messagesApi.preferred(request).lang.code}"
+              )
+            case _                                                        => throw new IllegalArgumentException("No form option is selected. Empty value for form.")
+          }
```
pascal22p, line 61: An exception? Are you sure?

Use a bad request with the form again and an error message

app/views/interstitial/SPPInterstitialView.scala.html
```diff
@@ -0,0 +1,68 @@
...
+                    value = Some("saBppAdvancePayment"),
+                    checked = formSaPaymentSelect("saBppWhatPaymentType").value.contains("saBppAdvancePayment")
+                )
+            )
+        )
+
+        <button id="continue-button" type="submit" class="govuk-button" data-module="govuk-button">
+            @messages("global.label.continue_button")
+        </button>
+    }
```
pascal22p, line 66: @govukButton(
            Button(
               id = Some("continue-button")
                element = Some("button"),
                content = Text(messages("global.label.continue_button"))
            )
        )

pascal22p (approved):

PR #935, opened 2025-09-15
DDCNL-10640: Adding new badges for app-store and play-store downloads

askar-makkunnath:


it/test/testUtils/IntegrationSpec.scala
```diff
@@ -294,7 +294,8 @@ trait IntegrationSpec
       |        {
       |            "pattern": "^/personal-account",
       |            "skinElement": "skinElement",
-      |            "isEnabled": true
+      |            "isEnabled": true,
+      |            "chatType": "loadWebChatContainer"
```
pascal22p, line 298: Use loadHMRCChatSkinElement

it/test/testUtils/IntegrationSpec.scala
```diff
@@ -340,7 +341,8 @@ trait IntegrationSpec
...
```
pascal22p, line 345: same here loadHMRCChatSkinElement

pascal22p (changes requested):

PR #991, opened 2026-03-19
[DDCNL-11708]: add GetMatchingFromCitizenDetailsToggle and fixed failing tests

klawani:
jira tickets: [DDCNL-11708](https://jira.tools.tax.service.gov.uk/browse/DDCNL-11708) 

- add GetMatchingFromCitizenDetailsToggle
- fixed failing tests


SandeepUmredkar:
Line 388 : "return None when connector returns a Json with no body"
still using the old toggle for mocking
Line 420: "return None when GetPersonFromCitizenDetailsToggle is disabled" in {
Name is wrong and also the toggle used in the test
it should be "return None when GetMatchingFromCitizenDetailsToggle is disabled"

There is one thing I observed in the current code and not new changes code
"return None when is GetPersonFromCitizenDetailsToggle disabled"
checks for 
verify(mockConnector, times(0)).getMatchingDetails(any())(any(), any())
it does not make sense, 
it should actually check for
verify(mockConnector, times(0)).personDetails(any(), any())(any(), any(), any())

SandeepUmredkar (commented):
Line 388 : "return None when connector returns a Json with no body"
still using the old toggle for mocking
Line 420: "return None when GetPersonFromCitizenDetailsToggle is disabled" in {
Name is wrong and also the toggle used in the test
it should be "return None when GetMatchingFromCitizenDetailsToggle is disabled"

There is one thing I observed in the current code and not new changes code
"return None when is GetPersonFromCitizenDetailsToggle disabled"
checks for
verify(mockConnector, times(0)).getMatchingDetails(any())(any(), any())
it does not make sense,
it should actually check for
verify(mockConnector, times(0)).personDetails(any(), any())(any(), any(), any())

SandeepUmredkar (approved):

dannybassi (approved):

PR #1015, opened 2026-05-21
DDCNL-11984 Removing old home page code, fandf banner code

Nj147:


it/test/controllers/TimeoutsISpec.scala
```diff
@@ -208,7 +208,7 @@ class TimeoutsISpec extends IntegrationSpec {
         Messages("label.this.section.is") + " " + Messages("label.account_home")
 
       contentAsString(result).contains(
-        Messages("label.transfer_part_of_your_personal_allowance_to_your_partner_")
+        Messages("Marriage Allowance")
```
SandeepUmredkar: If we are comparing strings directly then Messages is not needed

SandeepUmredkar: oldHomePageChangesBannerView is possibly not needed anymore and so these keys below are not needed

old_home_page_banner.text=We’re making changes to your personal tax account homepage.
old_home_page_banner.link_text=Read more about what’s changing
new_home_page_banner.text=Your personal tax account homepage has been updated.
new_home_page_banner.link_text=Find out what’s changed

and also this methods needs to  change

def getHomePageChangesBannerContent(
    newDesign: Boolean
  )(implicit ec: ExecutionContext, messages: Messages): Future[Option[Html]] =
    featureFlagService.get(HomePageChangesBannerToggle).map {
      case toggle if toggle.isEnabled =>
        if (newDesign) Some(newHomePageChangesBannerView()) else Some(oldHomePageChangesBannerView())
      case _                          => None
    }


these files does not seem to be used anywhere else
    views/tags/card.scala.html
    tags/cardHelper.scala.html
    tags/cardNewsAndUpdates.scala.html
   assets/stylesheets/partials/_home.scss
   assets/stylesheets/_flexbox.scss 
   assets/javascripts/card.js

   label.a_payment_to_help_with_the_cost_of_bringing_up_children label.mtd_for_itsa label.you_currently_transfer_part_of_your_personal_allowance_to_your_partner
   label.your_national_insurance_and_state_pension

these keys are no longer used now

Nj147: Removed js and css files, kept new home page banner as it's still toggled on in prod, removed old banner and messages. Some of those messages at the bottom are still used but removed the ones that aren't

SandeepUmredkar (dismissed):

SandeepUmredkar (approved):

PR #1016, opened 2026-05-21
PTAD-104 Add HMRC card container component

olakunlekuye2:
Add CardContainer and HMRCCard view models with accessibility validation.
Render empty, single-card, and multi-card states using the HMRC card list markup.
Add unit and accessibility coverage for rendering, list semantics, and accessible names.


app/viewmodels/CardContainerModel.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package viewmodels
+
+import play.twirl.api.{Html, HtmlFormat}
+
+final case class CardContainerModel(
+  emptyView: Html,
+  header: Option[CardContainerModel.Header] = None,
+  cards: Seq[HMRCCardModel] = Seq.empty,
```
owen-gilfellon, line 24: Will need to update this to use the HmrcCardModel from PTAD-100, once it's merged in... We can probably create some tickets for the next sprint to do integration, if we don't have time in this

app/views/tags/CardContainer.scala.html
```diff
@@ -0,0 +1,49 @@
...
+@if(model.normalizedHeader.isDefined) {
+  <@headingTag id="@headerIdValue" class="govuk-heading-m">
+    @model.normalizedHeader.get
+  </@headingTag>
+}
+
+@if(model.cards.isEmpty) {
+  @model.emptyView
+} else if(model.cards.length == 1) {
+  <div class="hmrc-card">
```
owen-gilfellon, line 36: Once PTAD-100 is merged in we can use HmrcCard.scala.html as a component in this file to render out the cards

owen-gilfellon (changes requested):
Looks good! A couple of changes required once PTAD-100 is merged in. If we can't get them done this sprint, we can raise a ticket to get it done in the next. The PR should target ptap-feature, though, not main

GPPWalton (approved):

owen-gilfellon (approved):

PR #1020, opened 2026-06-01
DDCNL-11987 - Refactor ApplicationController

venugopal-poli:


app/controllers/ApplicationController.scala
```diff
@@ -35,10 +35,6 @@ class ApplicationController @Inject() (
   val identityVerificationFrontendService: IdentityVerificationFrontendService,
   cc: MessagesControllerComponents,
   successView: SuccessView,
-  cannotConfirmIdentityView: CannotConfirmIdentityView,
```
Nj147, line 38: If these views aren't used elsewhere you can delete them

app/controllers/ApplicationController.scala
```diff
@@ -95,34 +89,16 @@ class ApplicationController @Inject() (
...
-
-            case LockedOut =>
-              logErrorMessage(LockedOut.toString)
-              Unauthorized(lockedOutView())
-
-            case Timeout =>
-              logErrorMessage(Timeout.toString)
-              Unauthorized(timeOutView())
+              logger.error(
+                s"PreconditionFailed response from IdentityVerificationFrontendService for journeyId: $journeyId. " +
```
SandeepUmredkar: journeyid is some, use jid

app/controllers/ApplicationController.scala
```diff
@@ -95,34 +89,16 @@ class ApplicationController @Inject() (
...
+                  "This outcome should not occur because the user should not have been able to enter the service. " +
+                  "Investigate the identity verification journey for this journeyId."
+              )
+              InternalServerError(technicalIssuesView(retryUrl))
 
             case TechnicalIssue =>
-              logger.warn(s"TechnicalIssue response from identityVerificationFrontendService")
+              logger.warn(s"TechnicalIssue response from IdentityVerificationFrontendService for journeyId: $journeyId")
```
SandeepUmredkar: same here use jid

app/controllers/ApplicationController.scala
```diff
@@ -95,34 +89,16 @@ class ApplicationController @Inject() (
...
-              Unauthorized(cannotConfirmIdentityView(retryUrl))
-
-            case LockedOut =>
-              logErrorMessage(LockedOut.toString)
-              Unauthorized(lockedOutView())
-
-            case Timeout =>
-              logErrorMessage(Timeout.toString)
-              Unauthorized(timeOutView())
+              logger.error(
```
SandeepUmredkar, line 114: Need to confirm if             case InsufficientEvidence => Redirect(routes.SelfAssessmentController.ivExemptLandingPage(continueUrl))
, is needed or not, because its not error I think

SandeepUmredkar:
Remove objects no longer needed in IdentityVerificationResponse

label.you_have_not_given_us_enough_information
label.you_have_tried_to_confirm_your_identity_
label.try_again_in_24_hours
label.you_have_been_signed_out

these keys are no longer used because you deleted views.

app/controllers/ApplicationController.scala
venugopal-poli: Deleted unused views and message labels which used in those views.

app/controllers/ApplicationController.scala
venugopal-poli: Addressed

app/controllers/ApplicationController.scala
venugopal-poli: Addressed.

venugopal-poli:
> Remove objects no longer needed in IdentityVerificationResponse
> 
> label.you_have_not_given_us_enough_information label.you_have_tried_to_confirm_your_identity_ label.try_again_in_24_hours label.you_have_been_signed_out
> 
> these keys are no longer used because you deleted views.

Resolved

app/controllers/ApplicationController.scala
venugopal-poli: Reverted back this case.

Nj147 (dismissed):

app/controllers/ApplicationController.scala
```diff
@@ -88,54 +85,38 @@ class ApplicationController @Inject() (
     continueUrl: Option[RedirectUrl],
     safeUrl: String
   )(implicit request: Request[AnyContent]): Future[Result] =
-    journeyId match {
-      case Some(jid) =>
+    (journeyId, request.session.get(SessionKeys.authToken)) match {
```
pascal22p: Won't work. You can spoof a fake token. You need to call auth so that the token is verified.

pascal22p (changes requested):

app/controllers/ApplicationController.scala
venugopal-poli: Have addressed.

pascal22p (approved):

PR #1029, opened 2026-06-15
PTAD-142 Add Tasks and Activities tab content

olakunlekuye2:
Implements PTAD-142 - PTAP Build: Tabular front end - Tasks and Activities content
Adds TabEnum, TabContentService, updates PtapHomeViewModel/HomeController/PtapHomeView
Renders tab content using SecondaryNav and CardContainer with Cards

### DIFF
--- a/app/controllers/HomeController.scala
+++ b/app/controllers/HomeController.scala
@@ -87,31 +87,27 @@ class HomeController @Inject() (
       enforceInterrupts {
         val fBreathingSpaceIndicator = breathingSpaceService.getBreathingSpaceIndicator(nino)
         val fEitherPersonDetails     = citizenDetailsService.personDetails(nino).value
-        val fTaskCards               = tabContentService.getTaskCards
-        val fTabCards                = currentTab match {
-          case Task     => fTaskCards
-          case Activity => tabContentService.getActivityCards
-          case _        => Future.successful(Seq.empty)
-        }
+        val fTabContentCards         = tabContentService.getTaskAndTabCards(currentTab)
 
         for {
           breathingSpaceIndicator <- fBreathingSpaceIndicator
           eitherPersonDetails     <- fEitherPersonDetails
           alertBannerContent      <- alertBannerHelper.getContent(eitherPersonDetails.toOption.flatten)
-          taskCards               <- fTaskCards
-          tabCards                <- fTabCards
+          tabContentCards         <- fTabContentCards
         } yield {
           val personDetailsOpt = eitherPersonDetails.toOption.flatten
           val nameToDisplay    = Some(personalDetailsNameOrDefault(personDetailsOpt))
 
-          val taskCount               = taskCards.size
+          val taskCount               = tabContentCards.taskCount
           implicit val msgs: Messages = messagesApi.preferred(request)
           val secondaryNav            = buildSecondaryNav(currentTab, taskCount)
-          val tabContent              = CardContainerModel(
-            emptyView = Html(""),
-            header = Some(msgs(currentTab.headingKey)),
-            cards = tabCards,
-            headerId = Some("tab-content-header")
+          val tabContent              = List(
+            CardContainerModel(
+              emptyView = Html(""),
+              header = Some(msgs(currentTab.headingKey)),
+              cards = tabContentCards.tabCards,
+              headerId = Some("tab-content-header")
+            )
           )
 
           Ok(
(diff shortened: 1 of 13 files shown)


GPPWalton (approved):
Looks good to me!

app/controllers/HomeController.scala
```diff
@@ -71,9 +74,11 @@ class HomeController @Inject() (
         }
       }
     }
-  private def personalisationHomePageTab(tab: String)(implicit request: UserRequest[AnyContent]): Future[Result] =
 
-    withValidTab(tab) { currentTab =>
+  private def personalisationHomePageTab(tab: String, allowTaskTab: Boolean = false)(implicit
```
owen-gilfellon: What is allowTaskTab used for? Is it needed?

app/controllers/HomeController.scala
```diff
@@ -83,36 +88,81 @@ class HomeController @Inject() (
...
+              cards = tabContentCards.tabCards,
+              headerId = Some("tab-content-header")
+            )
+          }.toList
+
           Ok(
             pTapHomeView(
               PtapHomeViewModel(
-                listOfTasks,
+                Seq.empty,
```
owen-gilfellon: If this isn't required (passed in via tabContent?), can it be removed?

app/controllers/HomeController.scala
```diff
@@ -83,36 +88,81 @@ class HomeController @Inject() (
...
-          listOfTasks             <- fListOfTasks
           eitherPersonDetails     <- fEitherPersonDetails
           alertBannerContent      <- alertBannerHelper.getContent(eitherPersonDetails.toOption.flatten)
+          tabContentCards         <- fTabContentCards
         } yield {
           val personDetailsOpt = eitherPersonDetails.toOption.flatten
           val nameToDisplay    = Some(personalDetailsNameOrDefault(personDetailsOpt))
 
+          val taskCount               = tabContentCards.taskCount
+          implicit val msgs: Messages = messagesApi.preferred(request)
```
owen-gilfellon: This is already provided implicitly, can be removed

app/controllers/HomeController.scala
```diff
@@ -160,23 +210,23 @@ class HomeController @Inject() (
...
     request: UserRequest[AnyContent]
   ): Future[Result] =
     val currentTab: Option[TabEnum] = tab match {
-      case Task.name     => Some(Task)
-      case Activity.name => Some(Activity)
-      case Tax.name      => Some(Tax)
-      case News.name     => Some(News)
-      case Support.name  => Some(Support)
-      case _             => None
+      case Task.name if allowTaskTab => Some(Task)
```
owen-gilfellon: is allowTaskTab needed?

owen-gilfellon (changes requested):
Looks good, just a couple of tweaks.

I noticed when I opened it in a browser - Now we're adding content, the spacing is incorrect under the secondary nav. In the prototype, I think `govuk-grid-row` is used to achieve the spacing... could you try adding them in to see if it corrects the issue?

app/controllers/HomeController.scala
olakunlekuye2: It wasn't needed. It was there to prevent /personal-account/your-tasks from working (since /personal-account already renders the Task tab via index), but both URLs produce identical content so there's no reason to block it. Removed allowTaskTab from personalisationHomePageTab, withValidTab, and the index call site. Also removed the associated test.

app/controllers/HomeController.scala
olakunlekuye2: Correct, tasks are now provided via tabContent through TabContentService, so the tasks field on PtapHomeViewModel was dead weight. Removed it from the model and the controller construction.

app/controllers/HomeController.scala
olakunlekuye2: Removed. Messages is already available implicitly in scope from the request.

app/controllers/HomeController.scala
olakunlekuye2: No , removed. Task is now always valid in withValidTab without the guard.

app/views/PtapHomeView.scala.html
```diff
@@ -68,7 +68,11 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-6">@viewModel.name.fold(messag
 
     @{SecondaryNav(viewModel.secondaryNav)}
 
-    @for(container <- viewModel.tabContent) {
-        @cardContainer(container)
-    }
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-full">
```
owen-gilfellon: *govuk-grid-column-two-thirds 🙂

owen-gilfellon (changes requested):

app/views/PtapHomeView.scala.html
olakunlekuye2: Updated to govuk-grid-column-two-thirds 👍

owen-gilfellon (approved):

PR #1032, opened 2026-06-17
PTAD-102: Add Support View Tab content

klawani:
jira ticket: [PTAD-102](https://jira.tools.tax.service.gov.uk/browse/PTAD-102)

Add English support content with URLs
Add placeholders for Welsh translation
Add Unit testings


GPPWalton (approved):

olakunlekuye2 (changes requested):
The external links in the linkList helper have target="_blank" but are missing rel="noopener noreferrer" — the ticket requires "correct security attributes" on new-tab links (Scenario 7). Check HmrcCard.scala.html line 27 for the existing pattern. Also, most external links don't include "(opens in new tab)" wording for the user — only the National Insurance one does, and the ticket asks for this to be clear to users in both languages.

olakunlekuye2 (changes requested):
Two typos in messages: "Paye As You Earn" should be "Pay", and "Mariage Allowance" should be "Marriage".

klawani:
> Two typos in messages: "Paye As You Earn" should be "Pay", and "Mariage Allowance" should be "Marriage".

DONE

klawani:
> The external links in the linkList helper have target="_blank" but are missing rel="noopener noreferrer" — the ticket requires "correct security attributes" on new-tab links (Scenario 7). Check HmrcCard.scala.html line 27 for the existing pattern. Also, most external links don't include "(opens in new tab)" wording for the user — only the National Insurance one does, and the ticket asks for this to be clear to users in both languages.

DONE

olakunlekuye2 (approved):

app/views/PtapHomeView.scala.html
```diff
@@ -73,4 +74,8 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-6">@viewModel.name.fold(messag
           TabModel(text = messages("ptap.support.uya.p6.sub"), href = Support.href(), current = Support == viewModel.currentTab),
         )
       ))}
+
+    @if(viewModel.currentTab == Support) {
+        @supportView()(messages)
```
owen-gilfellon: messages is implicit so doesn't need to be passed in directly

owen-gilfellon (changes requested):
I think we need to get PTAD-142 merged in first, and see how this code sits alongside that in PtapHomeView.

The formatting is off - we need to use govuk-grid-row etc to get the correct spacing / row widths - see PTAD-142

app/views/PtapHomeView.scala.html
klawani: Done

klawani:
> I think we need to get PTAD-142 merged in first, and see how this code sits alongside that in PtapHomeView.
> 
> The formatting is off - we need to use govuk-grid-row etc to get the correct spacing / row widths - see PTAD-142

Makes sense

olakunlekuye2 (approved):

PR #1034, opened 2026-06-19
PTAD-145 Tabular front end - Taxes and Benefits

olakunlekuye2:
Implemented the Taxes and Benefits tab content for the PTAP tabular front end.


GPPWalton (approved):
Looks good! The only thing I can see is the lack of bottom margin on the Secondary Nav, but I assume that the changes made in PTAD-142 will account for that, so it is all good!

owen-gilfellon (approved):

PR #1040, opened 2026-06-29
PTAD-176  NINO eligibility gate for new PTAP homepage design

olakunlekuye2:
Adds a configurable NINO eligibility check to the HomePagePersonalisationToggle gate
  The new PTAP design is shown only when the toggle is enabled and the user's last numeric NINO digit is in the configured rollout list
  No rollout band is hard-coded; the digit list is config-driven via feature.ptap-homepage.nino-rollout.last-numeric-digits with a safe empty default (no users enter PTAP unless app-config enables digits)


GPPWalton (approved):
Looks solid to me!

owen-gilfellon (changes requested):
Looks good, I'd probably add all the digits 0-9 to the configuration as a default while we're developing, so we don't have to manually edit while developing / editing each ticket.

owen-gilfellon (approved):

PR #1043, opened 2026-07-06
PTAD-172: PTAP UCD: Tabular front end - Default text for no tasks and activity tabs

GPPWalton:
Ticket: [PTAD-172](https://jira.tools.tax.service.gov.uk/browse/PTAD-172)

- Added new lines to the messages file for placeholder messages.
  - Added place holders for Welsh in interim before Welsh content is updated.
- empty() method added to TabEnum, which returns the defined placeholder text if present for that tab.
- Added unit test to PtapHomeViewSpec check rendering of the default text of 'Your tasks' tab and 'Recent activity' tab

### DIFF
--- a/app/viewmodels/PtapHomeViewModel.scala
+++ b/app/viewmodels/PtapHomeViewModel.scala
@@ -17,6 +17,7 @@
 package viewmodels
 
 import models.{MyService, OtherService}
+import play.api.i18n.Messages
 import play.twirl.api.Html
 
 enum TabEnum(val name: String, val cardContainerHeading: Option[String] = None):
@@ -26,9 +27,22 @@ enum TabEnum(val name: String, val cardContainerHeading: Option[String] = None):
   case News extends TabEnum("hmrc-news")
   case Support extends TabEnum("support")
 
-  def href(): String = this match {
+  def href(): String                             = this match {
     case Task => "/personal-account"
     case tab  => s"/personal-account/${tab.name}"
+  }
+  def empty()(implicit messages: Messages): Html = this match {
+    case Task =>
+      Html(s"""
+                <div class="govuk-inset-text">
+                  <p class="govuk-body">${messages("ptap.tasks.uya.default.l1")}</p>
+                  <p class="govuk-body">${messages("ptap.tasks.uya.default.l2.1")} <a href="${Tax
+          .href()}" class="govuk-link">${messages("ptap.tasks.uya.default.l2.2")}</a> ${messages(
+          "ptap.tasks.uya.default.l2.3"
+        )}</p>
+                </div>
+              """)
+    case _    => Html("")
   }
 
 final case class PtapNewsAndUpdates(content: Html) extends AnyVal
--- a/app/viewmodels/PtapHomeViewModel.scala
+++ b/app/viewmodels/PtapHomeViewModel.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 HM Revenue & Customs
+ * Copyright 2026 HM Revenue & Customs
  *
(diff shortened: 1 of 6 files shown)


app/viewmodels/PtapHomeViewModel.scala
```diff
@@ -26,10 +27,37 @@ enum TabEnum(val name: String, val cardContainerHeading: Option[String] = None):
   case News extends TabEnum("hmrc-news")
   case Support extends TabEnum("support")
 
-  def href(): String = this match {
+  def href(): String                             = this match {
     case Task => "/personal-account"
     case tab  => s"/personal-account/${tab.name}"
   }
+  def empty()(implicit messages: Messages): Html = this match {
+    case Task     =>
```
owen-gilfellon: Not keen on having all this HTML outside a view class... move to a scala.html file?
A a lot of the HTML is duplicated... if you renamed the keys in the message file you could do something like:
``ptap.${currentTab.name()}.uya.default.l1``
and reuse much of the code for each

conf/messages
```diff
@@ -859,6 +859,20 @@ label.mtdit.claim.question=Do you want to add Making Tax Digital for Income Tax
 label.mtdit.claim.error=Select yes to add Making Tax Digital for Income Tax to your personal tax account
 head.label.mtdit.claim.error=Select yes to add Making Tax Digital for Income Tax to your personal tax account
 
+#************************************************************
+# Default case (no tasks)
+#************************************************************
+ptap.tasks.uya.default.l1=This page shows refunds and tax you owe.
```
owen-gilfellon: Make message keys more descriptive, so they give some indication about what they refer to in the code

test/views/html/PtapHomeViewSpec.scala
```diff
@@ -285,5 +279,46 @@ class PtapHomeViewSpec extends ViewSpec {
...
+        .select("div.govuk-grid-column-two-thirds")
+        .select("div.govuk-inset-text")
+      placeholder_text.size mustBe 1
+      placeholder_text.select("p.govuk-body").size mustBe 2
+
+      val first_line  = placeholder_text.select("p.govuk-body").asList().get(0)
+      val second_line = placeholder_text.select("p.govuk-body").asList().get(1)
+      first_line.text()                          must include(messages("ptap.tasks.uya.default.l1"))
+      second_line
+        .text() mustBe s"${messages("ptap.tasks.uya.default.l2.1")} ${messages("ptap.tasks.uya.default.l2.2")} ${messages("ptap.tasks.uya.default.l2.3")}"
```
owen-gilfellon: Tests checking against message keys are circular - if the message key isn't present in the messages file, I think the output of messages() is just the message key. So, these tests could pass, even if there wasn't a `ptap.tasks.uya.default.l2.2` in the messages file.

Instead, if checking content, check against the rendered string i.e. first_line.text() must include "This page shows refunds and tax you owe."

owen-gilfellon (changes requested):

test/views/html/PtapHomeViewSpec.scala
GPPWalton: Ah I don't know why I missed that, sorted! 👍

conf/messages
GPPWalton: Good shout!

app/viewmodels/PtapHomeViewModel.scala
GPPWalton: Happy to extract it out! Hmm, would that mean I'd need to rejig the CardContainer, since it accepts an emptyView Html? I'll have a think/play and figure it out.

owen-gilfellon (approved):

PR #1045, opened 2026-07-10
PTAD-203: Private Beta - Update default text behaviour for Tasks and Recent Activity tabs

GPPWalton:
Ticket: [PTAD-203](https://jira.tools.tax.service.gov.uk/browse/PTAD-203)

- Removed emptyView from CardContainer and replaced with defaultInset

### DIFF
--- a/app/views/tags/CardContainer.scala.html
+++ b/app/views/tags/CardContainer.scala.html
@@ -34,7 +34,7 @@
 
 @if(model.cards.length == 1) {
   @HmrcCard(model.cards.head)
-} else {
+} else if(!model.cards.isEmpty) {
   <ul class="hmrc-card__container"
       @listLabel.map { label => aria-label="@label" }
       @if(listLabel.isEmpty) { aria-labelledby="@headerIdValue" }>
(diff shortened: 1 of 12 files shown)


olakunlekuye2 (commented):
The permanent inset text behaviour is implemented, but I don’t think the ticket is fully complete yet. The 10-day task message still needs to appear under the task list when tasks are present, and the Welsh content needs to be finalised. I understand there may still be some back and forth on the Welsh content.

app/views/tags/CardContainer.scala.html
```diff
@@ -30,10 +30,9 @@
     @model.normalizedHeader.get
   </@headingTag>
 }
+@model.defaultInset
 
-@if(model.cards.isEmpty) {
-  @model.emptyView
-} else if(model.cards.length == 1) {
+@if(model.cards.length == 1) {
```
olakunlekuye2, line 35: This change means `cards.length == 0` now falls through to the `else` below and renders an empty `<ul class="hmrc-card__container">`.

For the no tasks/no recent activity state, we should render the inset only and no empty card list. Could we add an explicit `model.cards.isEmpty` branch here, and cover it with a test?

GPPWalton: Good catch, I'll make ammendments!

olakunlekuye2 (approved):
LGTM

owen-gilfellon (approved):

PR #1052, opened 2026-07-13
PTAD-206: review and update use of “HMRC Online” wording

klawani:
jira ticket: [PTAD-206](https://jira.tools.tax.service.gov.uk/browse/PTAD-206)

- review and update the use of “HMRC Online” wording
- review and update the use of "Personal tax account"


olakunlekuye2 (commented):
PTAD-206 changes some wording to “Personal tax account”, but conf/messages:845 still says Your current HMRC Online taxes and benefits. Is that an agreed retained use of HMRC Online, or should it be aligned too?

PR #1053, opened 2026-07-15
PTAD-214: Private Beta - Tabular design - Fix Welsh translation for content headers in Your Tasks and Recent Activity tabs

GPPWalton:
Ticket: [PTAD-214](https://jira.tools.tax.service.gov.uk/browse/PTAD-214)

- refactored TabEnum, removing the parameter cardContainerHeading and replacing it with a method of the same name, which uses messages to return the correct local translation
- added welsh content for *"Recent activity"* heading

### DIFF
--- a/app/viewmodels/PtapHomeViewModel.scala
+++ b/app/viewmodels/PtapHomeViewModel.scala
@@ -21,9 +21,9 @@ import views.html.components.TabDefaultInsetText
 import play.api.i18n.Messages
 import play.twirl.api.Html
 
-enum TabEnum(val name: String, val cardContainerHeading: Option[String] = None):
-  case Task extends TabEnum("your-tasks", Some("Your tasks"))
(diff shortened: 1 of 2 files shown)


olakunlekuye2 (approved):
LGTM

app/viewmodels/PtapHomeViewModel.scala
owen-gilfellon: English is still displayed - need to remove the cardContainerHeading val from TabEnum

owen-gilfellon (changes requested):

app/viewmodels/PtapHomeViewModel.scala
GPPWalton: Weird, I might have accidentally dropped it when fixing a conflict, Will sort it out!

owen-gilfellon (approved):

PR #1073, opened 2026-08-17
PTAD-185 Return Tasks for the Your Tasks tab

olakunlekuye2:
- Added `TasksAndActivitiesConnector` to call `pta-tasks-and-events`.
- Added `tasks-and-activities-service-toggle`, defaulted off for controlled rollout.
- Updated `TasksService` to:
  - keep existing task logic when the toggle is off
  - call the new service when the toggle is on
  - fall back to existing logic if the toggle cannot be retrieved
  - handle service errors and empty responses safely
- Maps the new service response into the existing `Task` view model so the frontend presentation remains unchanged.
- Added connector and service tests for success, empty response, errors, and toggle behaviour.


GPPWalton (approved):
Looks good!

PR #1077, opened 2026-08-20
PTAD-239  : read URLs from config 

olakunlekuye2:
Replace servicesConfig.baseUrl() construction with direct config reads from external-url.lepp.url.


klawani (approved):
LGTM

PR #1078, opened 2026-08-21
PTAD-239 : hange leppUrl from nested (lepp { url }) to flat format (leppUrl)

olakunlekuye2:
Change leppUrl from nested (lepp { url }) to flat format (leppUrl)
to match app-config-base convention. ConfigDecorator already reads
from external-url.leppUrl, so this aligns the local config accordingly.


klawani (approved):
Looks good

PR #1080, opened 2026-09-09
PTAD-292 Update Trusted helpers content

olakunlekuye2:
- Updated the Trusted helpers content under Profile and settings.
- Added the confirmed English and Welsh title, supporting text and CTA wording.
- Kept the existing Trusted helpers link destination unchanged.
- Added test coverage for the updated English and Welsh content.


GPPWalton (approved):
Looking good!

PR #1081, opened 2026-09-10
PTAD-297 Update Ask HMRC Online digital assistant URL

olakunlekuye2:
- Updated the Ask HMRC Online chat link to point to the Digital Assistant URL.
- Applied the same destination for English and Welsh support card content.
- Updated Support view tests to assert the new Digital Assistant href.


GPPWalton (approved):
Looks good!

PR #1082, opened 2026-09-10
PTAD-267: Build - Display SA registration tile in Other taxes and benefits

GPPWalton:
Ticket: [PTAD-267](https://jira.tools.tax.service.gov.uk/browse/PTAD-267)

- added new case to getOtherSelfAssessment match block for NonFilerSelfAssessmentUser.
- accompanying test added.


owen-gilfellon (commented):
That looks fine and matches the ticket, but I think we need to confirm / update the content - it looks like it assumes the user will already know if they need to register, whereas the interstitial page will point them to guidance to determine that

it/test/controllers/HomeControllerSelfAssessmentISpec.scala
```diff
@@ -285,7 +285,8 @@ class HomeControllerSelfAssessmentISpec extends IntegrationSpec {
 
       val result: Future[Result] = route(app, request).get
       httpStatus(result) mustBe OK
-      contentAsString(result).contains(Messages("label.self_assessment")) mustBe false
+      print(contentAsString(result))
```
MarjanMubarok, line 288: You probably want to remove that

test/services/HomePageServicesProviderSpec.scala
```diff
@@ -435,6 +437,32 @@ class HomePageServicesProviderSpec extends BaseSpec {
...
+      val result = service.getHomePageServices(isRedesign = true).futureValue
+
+      result.otherServices.map(_.title)    must contain("Self Assessment")
+      result.otherServices.map(_.link)     must contain("/personal-account/self-assessment-who-needs-to-register")
+      result.otherServices.map(_.hintText) must contain(
+        Some("Check how to register for Self Assessment if you need to send a tax return.")
+      )
+    }
+
+    // "return welsh content version of self assessment in otherServices for non filer user" in {
```
MarjanMubarok, line 453: Remove this code if its not going to be required. Also I believe the general pattern is that you don't actually need to unit test welsh content.

it/test/controllers/HomeControllerSelfAssessmentISpec.scala
GPPWalton: good catch, will sort it. 👍

test/services/HomePageServicesProviderSpec.scala
GPPWalton: Our tickets previously (and this one) require welsh content testing as part of the ACs. That being said, forgot to sort this, thanks for spotting it!

it/test/controllers/HomeControllerSelfAssessmentISpec.scala
klawani: Done

test/services/HomePageServicesProviderSpec.scala
owen-gilfellon: Commented out test is still here - remove or uncomment.

klawani: Done. Should be good now

app/services/HomePageServicesProvider.scala
```diff
@@ -202,6 +203,14 @@ class HomePageServicesProvider @Inject() (
                 hint = Some(messages("label.activate_your_self_assessment"))
               )
             )
+          case NonFilerSelfAssessmentUser                      =>
+            Some(
+              otherSaTile(
+                title = messages("label.self_assessment"),
+                linkUrl = controllers.interstitials.routes.InterstitialController.displaySaRegistrationPage.url,
+                hint = Option.when(isRedesign)(messages("label.other_services.sa.reg.card.hint"))
```
owen-gilfellon: I don't think we need the isRedesign here - the hint texts for self assessment are always shown (it was an ask from Paul and David)

owen-gilfellon (changes requested):

app/services/HomePageServicesProvider.scala
klawani: This has now been removed. Thanks

PR #1083, opened 2026-09-11
 PTAD-289: Add confidence level check for LEPP backend personalisation API

olakunlekuye2:
Only call the LEPP backend/personalisation API when user has confidence level 250 or above. For users with CL200, show the static LEPP link under 'Other taxes and benefits' without calling the backend. This prevents 401 errors when accessing the personalised endpoint.

- Modified LeppService.getLeppLink to check request.confidenceLevel.level
- If CL >= 250: call backend for personalised content
- If CL < 250: return static start URL without backend call
- Updated tests with CL200 and CL250 scenarios


owen-gilfellon:
The scenario specified in the ticket (CL < 250) should be displayed under Other Services, not Current Services.
You seem to have duplicated the original message keys for the other LEPP scenarios - this scenario should have its own content i.e. "Check if you are eligible for a low earner’s pension payment" for this scenario, vs "View and accept your low earner''s pension payment." for the others.

owen-gilfellon (approved):

